PR #1, opened 2019-01-08
Size class added

katlapinka:

### DIFF
--- a/utils/size.py
+++ b/utils/size.py
@@ -1,0 +1,40 @@
+import enum
+
+
+class Unit(enum.Enum):
+    Byte = 1
+    KiloByte = 1000
+    KibiByte = 1024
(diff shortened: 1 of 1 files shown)


utils/size.py
```diff
@@ -0,0 +1,40 @@
+import enum
+
+
+class Unit(enum.Enum):
+    Byte = 1
+    KiloByte = 1000
+    KibiByte = 1024
+    MegaByte = 1000 * 1000
```
KamilLepek: 1000 * KiloByte

utils/size.py
```diff
@@ -0,0 +1,40 @@
...
+    MebiByte = 1024 * 1024
```
KamilLepek: 1024 * KibiByte

PR #32, opened 2019-08-29
CAS init config creation

katlapinka:
Add methods allowing to build CAS init config using existing and running CAS configuration on DUT. 
Symlink class added to get disk by-id path.
Minor CAS/utils API fixes.

### DIFF
--- a/api/cas/init_config.py
+++ b/api/cas/init_config.py
@@ -1,0 +1,83 @@
+#
+# Copyright(c) 2019 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+from api.cas import casadm_parser
+from cas_configuration.cache_config import CacheMode
+from storage_devices.device import Device
+from test_tools import fs_utils
+from config.configuration import cas_init_config_path
+
+
+def create_init_config_from_running_configuration(load: bool = None, extra_flags=""):
+    cache_lines = []
+    core_lines = []
+    for cache in casadm_parser.get_caches():
+        cache_lines.append(CacheConfigLine(cache.cache_id,
+                                           cache.cache_device,
+                                           cache.get_cache_mode(),
+                                           load,
+                                           extra_flags))
+        for core in casadm_parser.get_cores(cache.cache_id):
+            core_lines.append(CoreConfigLine(cache.cache_id,
+                                             core.core_id,
+                                             core.core_device))
+    config_lines = []
+    create_default_init_config()
+    if len(cache_lines) > 0:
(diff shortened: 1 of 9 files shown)


api/cas/init_config.py
```diff
@@ -0,0 +1,84 @@
+#
+# Copyright(c) 2019 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+from api.cas import casadm_parser
+from cas_configuration.cache_config import CacheMode
+from config.configuration import cas_init_config_path
```
Deixx: is it intended to be the default path or some custom path?
for the former I would specify it explicitly in this file, for the latter I would allow to specify path for output file as an argument in the method below

api/cas/init_config.py
```diff
@@ -0,0 +1,84 @@
...
+                                           cache.get_cache_mode(),
+                                           load,
+                                           extra_flags))
+        for core in casadm_parser.get_cores(cache.cache_id):
+            core_lines.append(CoreConfigLine(cache.cache_id,
+                                             core.core_id,
+                                             core.core_device))
+    config_lines = []
+    cas_version = casadm_parser.get_casadm_version()
+    config_lines.append(f"version={'.'.join(str(x) for x in cas_version.release[0:3])}")
```
Deixx: replace with create_default_init_config()?

storage_devices/device.py
```diff
@@ -42,3 +42,9 @@ def unmount(self):
             TestProperties.LOGGER.info("Device is not mounted.")
         elif disk_utils.unmount(self):
             self.mount_point = None
+
+    def get_device_link(self, directory: str):
```
KamilLepek: Since everything in Linux is a file what do you think about making some sort of higher lvl abstraction to be a baseclass for both files/devices. Could have some common fields that are used in both. In addition what about dividing this method into two (one which returns all of symlinks, the other one to get just the first)

katlapinka: I added that second method for getting list of symlinks and I think that later the base class can be made as well, but for now it is not so important

api/cas/init_config.py
```diff
@@ -0,0 +1,83 @@
+#
+# Copyright(c) 2019 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+from api.cas import casadm_parser
+from cas_configuration.cache_config import CacheMode
+from storage_devices.device import Device
+from test_tools import fs_utils
+from config.configuration import cas_init_config_path
```
Deixx: add current path to example configuration file

PR #41, opened 2019-10-08
Dont discover partitions

mmichal10:


storage_devices/disk.py
```diff
@@ -23,54 +24,28 @@ class DiskType(Enum):
...
-                if is_part_line:
-                    part_id, part_type = self.__parse_partition_info(line)
-                    if part_type != disk_utils.PartitionType.extended:
-                        self.partitions.append(Partition(self, part_type, part_id))
-                elif line.startswith("Number"):
-                    is_part_line = True
+        self.umount_all_partitions()
+        if disk_utils.create_partition_table(self, part_table_type):
+            self.partition_table = part_table_type
+        self.partition_table = part_table_type
```
robertbaldyga: I believe we don't want to set it on error. BTW why does `create_partition_table()` return `False` instead of rising an exception?

test_package/conftest.py
```diff
@@ -144,12 +144,20 @@ def unmount_cas_devices():
             )
 
 
+def kill_all_io():
+    TestProperties.executor.execute("pkill --signal SIGKILL dd")
+    TestProperties.executor.execute("pkill --signal SIGKILL [Vv][Dd][Bb][Ee][Nn][Cc][Hh]*")
```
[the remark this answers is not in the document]
mmichal10: Yeah, i tried, but pkill said: "unrecognized option '--ignore-case'".

storage_devices/disk.py
mmichal10: `create_partition_table()` behavior should change after new logging mechanism will be introduced.

PR #49, opened 2019-10-22
Introduce "require_disk" pytest mark

robertbaldyga:


core/test_run_utils.py
```diff
@@ -17,7 +18,45 @@
...
+    cls.req_disks = dict(req_disks)
+    if len(req_disks) != len(cls.req_disks):
+        raise Exception("Disk name specified more than once!")
+
+
+TestRun.prepare = __prepare
+
+
+@classmethod
+def __setup_disks(cls):
```
Deixx: This comment will not be essential (because I don't really think we'll have such scenarios). A machine has 1 optane and 1 nand. We state required disks as {"cache1": [optane, nand], "cache2": [optane]} - it would be smart to assign nand as cache1, otherwise there will not be enough optanes ;)

core/test_run_utils.py
```diff
@@ -17,7 +18,45 @@
 
 
 @classmethod
-def __prepare(cls, dut_config):
+def __configure(cls, config):
```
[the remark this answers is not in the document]
robertbaldyga: In class method it's mandatory anyway.

core/test_run_utils.py
robertbaldyga: Done.

PR #50, opened 2019-10-22
Handle '&' in sed

Deixx:

### DIFF
--- a/test_tools/fs_utils.py
+++ b/test_tools/fs_utils.py
@@ -132,45 +132,34 @@ def diff(file, other_file):
         return output.stderr
 
 
-# Any character other than whitespaces could be used as a sed separator.
-# The separator must not be used in pattern string, which is given as a sed parameter, so there is
-# a short list of possible separator characters, from which the first one not occuring in patterns
-# is selected
-def get_sed_separator(first_pattern, second_pattern=None):
-    sed_separators = ['/', ':', ';', ',', '.']
-    if second_pattern is None:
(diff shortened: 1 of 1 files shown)


test_tools/fs_utils.py
```diff
@@ -156,19 +156,19 @@ def insert_line_before_pattern(file, pattern, new_line):
...
-def replace_first_pattern_occurence(file, pattern, new_line):
+def replace_first_pattern_occurrence(file, pattern, new_line):
     separator = get_sed_separator(pattern, new_line)
     pattern = pattern.replace("'", "\\x27")
-    new_line = new_line.replace("'", "\\x27")
+    new_line = new_line.replace("'", "\\x27").replace("&", "\\&")
     cmd = f"sed -i '0,{separator}{pattern}{separator}s" \
-        f"{separator}{pattern}{separator}{new_line}{separator}' {file}"
+        f"{separator}{separator}{new_line}{separator}' {file}"
     return TestRun.executor.run_expect_success(cmd)
```
robertbaldyga: You can achieve the same in plain python with no sorcery:
```python
pat = re.compile(pattern)
with open(file, "r+") as f:
    lines = (l for l in f.readlines())
    f.truncate(0)
    f.writelines(itertools.takewhile(lambda l: not pat.search(l), lines))
    f.write(new_line)
    f.writelines(lines)
```

Deixx: with open will not work with remote execution

PR #56, opened 2019-10-30
Fix for logging iterations and add Log destructor

katlapinka:


log/html_log_manager.py
```diff
@@ -68,6 +68,7 @@ def end_iteration(self):
         self._main.end_iteration(self._current_log.get_result())
         self._log_setup.end_iteration(self._current_log.get_result())
         self._current_log = self._log_setup
+        self._log_iterations[self._config.get_iteration_id() - 1].iteration_closed = True
```
robertbaldyga: Isn't is the same as `self._log_iterations[-1].iteration_closed = True`? Current iteration is the last in the list anyway. We can simplify it even further by changing order of operations, as initially `self._current_log` contains reference to current iteration, so we can do:
```python
self._current_log.iteration_closed = True
self._current_log = self._log_setup
```

PR #57, opened 2019-10-30
Add rsync() function to executor

robertbaldyga:
This function can be used to copy files or directories from
controller to DUT.

### DIFF
--- a/connection/base_executor.py
+++ b/connection/base_executor.py
@@ -11,6 +11,9 @@ class BaseExecutor:
     def _execute(self, command, timeout):
         raise NotImplementedError()
 
+    def rsync(self, src, dst, delete, timeout):
+        raise NotImplementedError()
+
     def is_remote(self):
         return False
 
--- a/connection/ssh_executor.py
+++ b/connection/ssh_executor.py
@@ -4,6 +4,7 @@
 #
 
 import socket
+import subprocess
 from datetime import timedelta, datetime
 
 import paramiko
@@ -48,6 +49,18 @@ class SshExecutor(BaseExecutor):
                                   f" {self.ip}\n{e}")
 
         return Output(stdout.read(), stderr.read(), stdout.channel.recv_exit_status())
+
+    def rsync(self, src, dst, delete=False, timeout: timedelta = timedelta(seconds=30)):
+        options = []
+        if delete:
+            options.append("--delete")
+        subprocess.run(
+            f'sshpass -p "{self.password}"'
+            f'rsync -r {src} {self.user}@{self.ip}:{dst} {" ".join(options)}',
+            shell=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            timeout=timeout.total_seconds())
 
     def is_remote(self):
         return True
--- a/connection/dummy_executor.py
+++ b/connection/dummy_executor.py
@@ -9,3 +9,6 @@ from connection.base_executor import BaseExecutor
 class DummyExecutor(BaseExecutor):
     def _execute(self, command, timeout=None):
         print(command)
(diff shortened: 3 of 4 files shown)


connection/ssh_executor.py
```diff
@@ -9,6 +9,7 @@
 import paramiko
 
 from connection.base_executor import BaseExecutor
+from connection.local_executor import LocalExecutor
```
Deixx: I'm not fond of creating LocalExecutor in SshExecutor

connection/ssh_executor.py
```diff
@@ -49,6 +50,13 @@ def _execute(self, command, timeout):
 
         return Output(stdout.read(), stderr.read(), stdout.channel.recv_exit_status())
 
+    def copy(self, src, dst, timeout: timedelta = timedelta(seconds=30)):
+        local_executor = LocalExecutor()
+        local_executor._execute(
+            f'sshpass -p "{self.password}" rsync -r {src} {self.user}@{self.ip}:{dst}',
```
Deixx: sshpass is another package - you could try using 
sftp = TestRun.executor.ssh.open_sftp()
sftp.put(...)

connection/base_executor.py
```diff
@@ -11,6 +11,9 @@ class BaseExecutor:
     def _execute(self, command, timeout):
         raise NotImplementedError()
 
+    def copy(self, src, dst, timeout):
```
Deixx: I'd rather put this in separate util, not executor
Copying in dummy is unimportant, in local is implemented in fs_utils, the only thing we need is copying between controller and DUT, which sounds like a separate util

connection/ssh_executor.py
robertbaldyga: Unfortunately `put()` is able to copy only single file at a time, so we would need to implement procedure of recursive directory tree copy, which `rsync -r` gives us out of the box.

connection/base_executor.py
robertbaldyga: The problem is that this implementation should be chosen depending on current executor, so we would need another util which would be parametrized with executor class. So we would have implementation per executor outside the executor which is not very object oriented design.
fs_utils are designed to work with DUT fs, not to provide bridge between controller fs and DUT fs. And the API we want framework to provide should be exactly the same for both local and remote execution.
IMO the name **executor** is misleading. What we really need under the **executor** is a complete interface between controller and DUT, not only dummy util for running commands.

connection/ssh_executor.py
robertbaldyga: LocalExecutor is a handy shortcut for subprocess, but if we prefer to use the latter explicitly, I can change it.

connection/base_executor.py
```diff
@@ -11,6 +11,9 @@ class BaseExecutor:
     def _execute(self, command, timeout):
         raise NotImplementedError()
 
+    def rsync(self, src, dst, delete, timeout):
```
katlapinka: I'd rather name this method like "copy_to_dut" or "rsync_controller_with_dut", because there is the other method in utils, which is used to copy files over dut, so it could be misleading to have similar methods, which are doing different things. (I understand that this method in here is used to syncronize files between controller and dut, no matter if controller is local or remote, and the other method I've mentioned is used to copying files always on the same server)

robertbaldyga: As this method is operation performed on **executor** which is actual interface between controller and DUT I don't find it misleading. Moreover we plan to rename **executor** to **dut** in same point in future (as this name is more precise), so `dut.copy_to_dut()` looks kind of weird. However I think we can rename it to `rsync_to()` to point the direction, as for now the only supported direction is from controller to DUT.

PR #65, opened 2019-11-06
Added 'port' option to DUT YAML config and extended rsync options

tomaszrybicki:
Signed-off-by: Tomasz Rybicki <tomasz.rybicki@intel.com>

### DIFF
--- a/connection/local_executor.py
+++ b/connection/local_executor.py
@@ -23,10 +23,18 @@ class LocalExecutor(BaseExecutor):
                       completed_process.stderr,
                       completed_process.returncode)
 
-    def rsync(self, src, dst, delete=False, timeout: timedelta = timedelta(seconds=30)):
+    def rsync(self, src, dst, delete=False, symlinks=False, exclude_list=[],
+              timeout: timedelta = timedelta(seconds=30)):
         options = []
+
         if delete:
             options.append("--delete")
+        if symlinks:
+            options.append("--links")
+
+        for exclude in exclude_list:
+            options.append(f"--exclude {exclude}")
+
         subprocess.run(
             f'rsync -r {src} {dst} {" ".join(options)}',
             shell=True,
--- a/connection/ssh_executor.py
+++ b/connection/ssh_executor.py
@@ -19,6 +19,7 @@ class SshExecutor(BaseExecutor):
         self.ip = ip
         self.user = username
         self.password = password
+        self.port = port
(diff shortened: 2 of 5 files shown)


core/test_run_utils.py
```diff
@@ -83,11 +83,18 @@ def __setup(cls, dut_config):
             IP(dut_config['ip'])
         except ValueError:
             raise Exception("IP address from configuration file is in invalid format.")
+
+        if 'port' in dut_config:
```
mmichal10: I believe `port = dut_config['port']` if 'port' in dut_config else 22` is more pythonic.

core/test_run_utils.py
```diff
@@ -83,11 +83,15 @@ def __setup(cls, dut_config):
             IP(dut_config['ip'])
         except ValueError:
             raise Exception("IP address from configuration file is in invalid format.")
+
+        port = dut_config['port'] if 'port' in dut_config else 22
```
robertbaldyga: This could be:
```python
port = dut_config.get('port', 22)
```
or even better:
```python
dut_config.setdefault('port', 22)
```

connection/local_executor.py
```diff
@@ -23,10 +23,18 @@ def _execute(self, command, timeout):
...
-    def rsync(self, src, dst, delete=False, timeout: timedelta = timedelta(seconds=30)):
+    def rsync(self, src, dst, delete=False, symlinks=False, exclude_list=[],
+              timeout: timedelta = timedelta(seconds=30)):
         options = []
+
         if delete:
             options.append("--delete")
+        if symlinks:
+            options.append("--links")
+        if len(exclude_list) > 0:
```
robertbaldyga: This check is unnecessary. The `for` loop in that case will just run 0 times.

connection/ssh_executor.py
```diff
@@ -50,13 +51,22 @@ def _execute(self, command, timeout):
...
-    def rsync(self, src, dst, delete=False, timeout: timedelta = timedelta(seconds=30)):
+    def rsync(self, src, dst, delete=False, symlinks=False, exclude_list=[],
+              timeout: timedelta = timedelta(seconds=30)):
         options = []
+
         if delete:
             options.append("--delete")
+        if symlinks:
+            options.append("--links")
+        if len(exclude_list) > 0:
```
robertbaldyga: Again, this check is unnecessary.

PR #68, opened 2019-11-12
Added CmdException + minor fixes

Deixx:
Needs: https://github.com/Open-CAS/open-cas-linux/pull/186

### DIFF
--- a/connection/base_executor.py
+++ b/connection/base_executor.py
@@ -4,7 +4,9 @@
 #
 
 from datetime import timedelta
+
 from core.test_run import TestRun
+from test_utils.output import CmdException
 
 
 class BaseExecutor:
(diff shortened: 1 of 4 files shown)


connection/base_executor.py
```diff
@@ -45,13 +47,13 @@ def wait_cmd_finish(self, pid: int, timeout: timedelta = timedelta(minutes=30)):
     def run_expect_success(self, command):
         output = self.run(command)
         if output.exit_code != 0:
-            raise Exception(f"Exception occurred while trying to execute '{command}' command.\n"
-                            f"stdout: {output.stdout}\nstderr: {output.stderr}")
+            raise CmdException(f"Exception occurred while trying to execute '{command}' command.",
```
katlapinka: I'd prefer to use TestRun.exception("...") to create a proper log with valid message. Current implementation causes unexpected end of log file and the result is passed but it shouldn't. It would be good idea to handle exception types some way in TetstRun.exception method.

robertbaldyga: As now we handle exceptions properly, I think we can merge this patch as it is.

PR #70, opened 2019-11-19
 Fs_utils and file modules update

Ostrokrzew:
Create 'create_test_file' function in 'fs_utils' module.
Create 'FileProperties' class in 'file' module to read current file properties.

Signed-off-by: Slawomir_Jankowski <slawomir.jankowski@intel.com>

### DIFF
--- a/test_tools/fs_utils.py
+++ b/test_tools/fs_utils.py
@@ -266,3 +266,11 @@ def parse_ls_output(ls_output, dir_path=''):
         fs_item.modification_time = modification_time
         fs_items.append(fs_item)
     return fs_items
+
+
+def create_test_file(target_file_path, file_content="Test content"):
+    from test_utils.filesystem.file import File
+    file = File.create_file(target_file_path)
+    file.write(file_content)
+    file.refresh_item()
+    return file
--- a/test_utils/filesystem/file.py
+++ b/test_utils/filesystem/file.py
@@ -2,19 +2,19 @@
 # Copyright(c) 2019 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 #
+import ntpath
 
 from test_tools.dd import Dd
 from test_utils.filesystem.fs_item import FsItem
 from test_utils.size import Size
 from test_tools import fs_utils
 from core.test_run import TestRun
-from test_tools import fs_utils
-from test_utils.filesystem.fs_item import FsItem
 
 
 class File(FsItem):
     def __init__(self, full_path):
         FsItem.__init__(self, full_path)
+        self.md5 = self.md5sum()
 
     def compare(self, other_file):
         return fs_utils.compare(str(self), str(other_file))
@@ -35,6 +35,9 @@ class File(FsItem):
     def write(self, content, overwrite: bool = True):
         fs_utils.write_file(str(self), content, overwrite)
         self.refresh_item()
+
+    def get_properties(self):
+        return FileProperties(self)
 
     @staticmethod
     def create_file(path: str):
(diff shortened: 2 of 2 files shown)


test_tools/fs_utils.py
```diff
@@ -266,3 +266,36 @@ def parse_ls_output(ls_output, dir_path=''):
...
+    if size_before != file_after.size:
+        TestRun.LOGGER.error(f"Size before ({size_before}) and after ({file_after.size}) "
+                             f"are different.")
+
+
+def create_test_file(target_file_path):
+    create_file(target_file_path)
+    output = ls_item(f"{target_file_path}")
+    file = parse_ls_output(output)[0]
+    file.write("Test content")
```
mmichal10: create_file() returns File object, so there is no need to parse ls output

test_tools/fs_utils.py
```diff
@@ -266,3 +266,36 @@ def parse_ls_output(ls_output, dir_path=''):
...
+                             f"and after ({file_after.permissions.group}) are different.")
+    if permissions_before.other != file_after.permissions.other:
+        TestRun.LOGGER.error(f"Other permissions before ({permissions_before.other}) "
+                             f"and after ({file_after.permissions.other}) are different.")
+    if size_before != file_after.size:
+        TestRun.LOGGER.error(f"Size before ({size_before}) and after ({file_after.size}) "
+                             f"are different.")
+
+
+def create_test_file(target_file_path):
```
mmichal10: Add optional param to allow caller set file padding.

test_tools/fs_utils.py
```diff
@@ -266,3 +266,36 @@ def parse_ls_output(ls_output, dir_path=''):
...
+    create_file(target_file_path)
+    output = ls_item(f"{target_file_path}")
+    file = parse_ls_output(output)[0]
+    file.write("Test content")
+    md5_before_load = file.md5sum()
+    size_before_load = file.size
+    permissions_before_load = file.permissions
```
mmichal10: If caller need those infos, he can retrieve them by himself, you don't have to return them.

test_tools/fs_utils.py
Ostrokrzew: done

test_tools/fs_utils.py
Ostrokrzew: good idea, done

test_tools/fs_utils.py
Ostrokrzew: done

test_tools/fs_utils.py
```diff
@@ -266,3 +266,35 @@ def parse_ls_output(ls_output, dir_path=''):
         fs_item.modification_time = modification_time
         fs_items.append(fs_item)
     return fs_items
+
+
+def compare_files(file):
+    check_files(file, file.size, file.permissions, file.md5sum())
```
robertbaldyga: So it just compares file properties with itself at the same moment? What is use case of this function?

BTW this function is missnamed. I'd expect that `compare_names()` takes two files as arguments and compares them, which here is clearly not the case as the file is compared with itself.

Ostrokrzew: This function takes file as and argument, reads its properties, then refreshes file's status, one more time reads its refreshed properties and finally compares them.
You're right, name of this function would be changed.

robertbaldyga: Oh, I see. So IMO having in class `File` (which obviously represents actual file) fields like md5 that may go out of sync and need calling special method for updating them is a terrible design decision. It would be much better to have class `FileProperties` representing set of properties at given moment in time that would be produced by `File`, so that `File` itself wouldn't have any expirable state. The `FileProperties` class would then have proper comparison operators and getters, producing intuitive and idiomatic API.

Ostrokrzew: Hope I done the work, but please, check it.

PR #74, opened 2019-11-26
Add method for running FIO in background

rafalste:

### DIFF
--- a/test_tools/fio/fio.py
+++ b/test_tools/fio/fio.py
@@ -54,17 +54,24 @@ class Fio:
         return datetime.timedelta(seconds=total_time)
 
     def run(self, timeout: datetime.timedelta = None):
-        if not self.is_installed():
-            self.install()
-
         if timeout is None:
             timeout = self.calculate_timeout()
+
+        self.prepare_run()
+        return self.executor.run(str(self), timeout)
+
+    def run_in_background(self):
+        self.prepare_run()
+        return self.executor.run_in_background(str(self))
+
+    def prepare_run(self):
+        if not self.is_installed():
+            self.install()
 
         if len(self.jobs) > 0:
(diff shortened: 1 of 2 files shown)


test_tools/fio/fio.py
```diff
@@ -54,6 +54,14 @@ def calculate_timeout(self):
         return datetime.timedelta(seconds=total_time)
 
     def run(self, timeout: datetime.timedelta = None):
+        timeout = self.prepare_run(timeout)
```
robertbaldyga: Passing timeout back and forth looks more confusing than useful. It would be more readable if `prepare_run()` had no argument `timeout`, and calculating `timeout`, if not specified, were done in `run()`.

Btw it can be simple like this:
```python
timeout = timeout or self.calculate_timeout()
```

test_tools/fio/fio_param.py
```diff
@@ -266,6 +266,12 @@ def run(self):
         output = self.command_executor.run(f"cat {self.fio.fio_file}")
         return self.get_results(output.stdout)
 
+    def run_in_background(self):
+        self.fio.base_cmd_parameters.set_flags("group_reporting")
+        if "per_job_logs" in self.fio.global_cmd_parameters.command_param_dict.keys():
```
robertbaldyga: In python3 you can check if key exists in dictionary without using method `keys()`.

This:
```python
if 'key' in dictionary.keys():
   ...
```
is functionally equivalent to:
```python
if 'key' in dictionary:
   ...
```

rafalste: done

test_tools/fio/fio.py
rafalste: done... except of `timeout = timeout or self.calculate_timeout()`
I think the old version is more readable.

PR #76, opened 2019-11-26
Improve handling of process killing ('open-cas-linux' + 'test-framework' changes)

rafalste:

### DIFF
--- a/connection/base_executor.py
+++ b/connection/base_executor.py
@@ -3,6 +3,8 @@
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 #
 
+import time
+
 from datetime import timedelta
 from core.test_run import TestRun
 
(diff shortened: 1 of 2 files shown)


connection/base_executor.py
```diff
@@ -42,6 +44,13 @@ def run_in_background(self, command):
     def wait_cmd_finish(self, pid: int, timeout: timedelta = timedelta(minutes=30)):
         self.run(f"tail --pid={pid} -f /dev/null", timeout)
 
+    def kill_process(self, pid: int):
+        self.run(f"kill -s SIGTERM {pid}")
+        time.sleep(3)
+        check_process_killed = TestRun.executor.run(f"ps -p {pid}")
+        if check_process_killed.exit_code == 0:
```
robertbaldyga: We don't need to bother checking, we can send SIGKILL anyway (like in `kill_all_io()`). Killing potentialy dead processes for sake of simplicity is quite ok.

rafalste: Ok, I'll do as you said... but just to explain myself:
in kill_all_io() there is a 'killall -q' which calls process by name and suppress all error messages and in this case we can kill process that not exist without an error message. On the other hand we have 'kill' that calls process by pid and complains when we try to kill non-existent process.

robertbaldyga: I see `kill` doesn't have `-q` option, but we can achieve the same effect by redirecting stdout/stderr to `/dev/null`:
```
kill -s SIGKILL {pid} &>/dev/null
```

rafalste: done + amendments from @mmichal10 :
- comment on using TERM before KILL signal
- check if kill_all_io() did his job

PR #78, opened 2019-11-27
Wrap device mapper and implement ErrorDevice

imjfckm:

### DIFF
--- a/test_tools/device_mapper.py
+++ b/test_tools/device_mapper.py
@@ -1,0 +1,328 @@
+#
+# Copyright(c) 2019 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+from enum import Enum
+
+from test_utils.linux_command import LinuxCommand
+from test_utils.size import Size, Unit
+from storage_devices.device import Device
+from core.test_run import TestRun
+from test_utils.output import Output
+
+
+class DmTarget(Enum):
+    # Fill argument types for other targets if you need them
+    LINEAR = (str, int)
+    STRIPED = (int, int, list)
+    ERROR = ()
+    ZERO = ()
+    CRYPT = ()
+    DELAY = (str, int, int, str, int, int)
+    FLAKEY = (str, int, int, int)
+    MIRROR = ()
+    MULTIPATH = ()
+    RAID = ()
+    SNAPSHOT = ()
+
+    def __str__(self):
+        return self.name.lower()
+
+
+class DmTable:
+    class TableEntry:
+        pass
+
+
+class DmTable:
+    class TableEntry:
+        def __init__(self, offset: int, length: int, target: DmTarget, *params):
+            self.offset = int(offset)
+            self.length = int(length)
+            self.target = DmTarget(target)
+            self.params = list(params)
+            self.validate()
+
+        def validate(self):
+            if self.target.value:
+                for i in range(len(self.params)):
+                    try:
+                        self.params[i] = self.target.value[i](self.params[i])
+                    except IndexError:
+                        raise ValueError("invalid dm target parameter")
+
+        def __str__(self):
+            ret = f"{self.offset} {self.length} {self.target}"
+            for param in self.params:
+                ret += f" {param}"
+
+            return ret
+
+    def __init__(self):
+        self.table = []
+
+    @classmethod
+    def uniform_error_table(
+        cls, start_lba: int, stop_lba: int, num_error_zones: int, error_zone_size: Size
+    ):
+        table = cls()
+        increment = (stop_lba - start_lba) // num_error_zones
+
+        for zone_start in range(start_lba, stop_lba, increment):
+            table.add_entry(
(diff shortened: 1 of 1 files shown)


test_tools/device_mapper.py
```diff
@@ -0,0 +1,353 @@
...
+
+        output += "ENDTABLE"
+
+        return output
+
+
+class DeviceMapper(LinuxCommand):
+    @classmethod
+    def remove_all(cls, force=True):
+        TestRun.LOGGER.info(f"Removing all device mapper devices")
```
micrakow: Is 'f' needed if no argument is being passed to build a string?

test_tools/device_mapper.py
```diff
@@ -0,0 +1,353 @@
...
+        result = self.mapper.remove()
+
+        return result
+
+    def change_table(self, table: DmTable, permanent=True):
+        if self.active:
+            result = self.mapper.suspend()
+            if result.exit_code:
+                TestRun.LOGGER.exception(
+                    f"couldn't suspend error device '{self.name}'!"
```
micrakow: General: how about printing err code as well?

test_tools/device_mapper.py
```diff
@@ -0,0 +1,353 @@
...
+
+    def suspend(self):
+        if not self.active:
+            TestRun.LOGGER.error(
+                f"cannot suspend error device '{self.name}'! It's already running"
+            )
+            return Output("", "", 1)
+
+        result = self.mapper.suspend()
+        if not result.exit_code:
```
micrakow: Is '0' an expected exit code in case of error?

test_tools/device_mapper.py
```diff
@@ -0,0 +1,353 @@
...
+
+    def resume(self):
+        if self.active:
+            TestRun.LOGGER.error(
+                f"cannot resume error device '{self.name}'! It's already running"
+            )
+            return Output("", "", 1)
+
+        result = self.mapper.resume()
+        if not result.exit_code:
```
micrakow: Is '0' an expected exit code in case of error?

test_tools/device_mapper.py
```diff
@@ -0,0 +1,353 @@
...
+    def reload(self, table: DmTable):
+        table.validate()
+        TestRun.LOGGER.info(f"Reloading table for device mapper device '{self.name}'")
+
+        for entry in table.table:
+            TestRun.LOGGER.debug(f"{entry}")
+
+        return TestRun.executor.run(
+            f"{self.command_name} reload {self.name} {table.to_heredoc()}"
+        )
```
robertbaldyga: Why not just do `run_expect_success()` instead of returning output which is then parsed the same way as `run_expect_success()` does?

test_tools/device_mapper.py
```diff
@@ -0,0 +1,353 @@
...
+        self.mapper = DeviceMapper(name)
+        self.name = name
+        self.table = DmTable.passthrough_table(base_device) if not table else table
+        self.active = False
+        self.start()
+
+    @property
+    def system_path(self):
+        if self.active:
+            output = TestRun.executor.run(f"realpath {self.mapper.get_path()}")
```
robertbaldyga: Ditto.

test_tools/device_mapper.py
imjfckm: Oh, didn't know we had that! Fixed and removed all redundant error handling

test_tools/device_mapper.py
imjfckm: Fixed

test_tools/device_mapper.py
imjfckm: Fixed

test_tools/device_mapper.py
imjfckm: Removed error handling and left only `run_expect_success` - the caller can handle exceptions if he wants to

test_tools/device_mapper.py
imjfckm: Fixed (removed)

test_tools/device_mapper.py
imjfckm: Same

test_tools/device_mapper.py
```diff
@@ -0,0 +1,325 @@
...
+                raise ValueError(
+                    f"dm table should not have any holes or overlaps: {e1} -> {e2}"
+                )
+
+    def get_size(self):
+        self.table.sort(key=lambda entry: entry.offset)
+
+        return Size(self.table[-1].offset + self.table[-1].length, Unit.Blocks512)
+
+    def to_heredoc(self):
```
robertbaldyga: IMO this heredoc is implementation detail of create/reload command invocation, not the `DmTable` itself. It would be clearer if `DmTable` had simple `__str__()` method returning only the table (without heredoc), and the heredoc part were appended during command invocation.

imjfckm: True that, changed

PR #82, opened 2019-11-29
size: add 'align' method

micrakow:


test_utils/size.py
```diff
@@ -122,6 +122,11 @@ def is_zero(self):
         else:
             return False
 
+    def align(self, alignment):
+        if alignment % 2:
```
robertbaldyga: This is not proper check for power of 2.

PR #83, opened 2019-12-06
Added timeout option to run_in_background method

tomaszrybicki:
Signed-off-by: Tomasz Rybicki <tomasz.rybicki@intel.com>

### DIFF
--- a/connection/base_executor.py
+++ b/connection/base_executor.py
@@ -36,9 +36,9 @@ class BaseExecutor:
         TestRun.LOGGER.write_output_to_command_log(output, command_id)
         return output
 
-    def run_in_background(self, command):
+    def run_in_background(self, command, timeout: timedelta = timedelta(minutes=30)):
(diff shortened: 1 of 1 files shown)


connection/base_executor.py
```diff
@@ -36,9 +36,9 @@ def run(self, command, timeout: timedelta = timedelta(minutes=30)):
         TestRun.LOGGER.write_output_to_command_log(output, command_id)
         return output
 
-    def run_in_background(self, command):
+    def run_in_background(self, command, timeout: timedelta = timedelta(minutes=30)):
         command += "&> /dev/null &echo $!"
-        output = self.run(command)
+        output = self.run(command, timeout)
```
robertbaldyga: This won't work as you may expect. This function just runs command in the background and it returns immediately. The state of the background process isn't tracked so the timeout doesn't apply to it. What you probably need is command line utility named `timeout`.

PR #89, opened 2019-12-10
Expose verification options for FIO

imjfckm:


test_tools/fio/fio_param.py
```diff
@@ -228,6 +228,10 @@ def verification_with_pattern(self, pattern=None):
     def verify(self, value: VerifyMethod):
         return self.set_param('verify', value.name)
 
+    def verify_pattern(self, pattern = None):
+        pattern = pattern if pattern else self.get_verification_pattern()
+        return self.set_param('verify_pattern', pattern)
```
robertbaldyga: This is shorter and pythonic:
```python
return self.set_param('verify_pattern', pattern or self.get_verification_pattern())
```

imjfckm: Pythonized

PR #91, opened 2019-12-17
Add method to base executor

Ostrokrzew:
Add method to run commands as other user and as other user with 'sudo'.

Signed-off-by: Ostrokrzew <slawomir.jankowski@intel.com>

### DIFF
--- a/connection/base_executor.py
+++ b/connection/base_executor.py
@@ -43,6 +43,16 @@ class BaseExecutor:
         if output is not None:
             return int(output.stdout)
 
+    def run_as_other_user(self, command, user_name: str, sudo: bool = False):
+        prefix = f'sudo -u {user_name}'
+        if sudo:
+            command = 'sudo ' + command
+        command = f'{prefix} {command}'
+        output = self.run(command)
+        if output.exit_code != 0:
+            raise CmdException("Must be run as root.", output)
+        return output
+
     def wait_cmd_finish(self, pid: int, timeout: timedelta = timedelta(minutes=30)):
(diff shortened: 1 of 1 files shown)


test_utils/linux_command.py
```diff
@@ -23,6 +23,12 @@ def run(self):
     def run_in_background(self):
         return self.command_executor.run_in_background(str(self))
 
+    def run_as_other_user(self, user_name: str):
+        return self.command_executor.run_as_other_user(str(self), user_name)
+
+    def run_as_other_user_sudo(self, user_name: str):
+        return self.command_executor.run_as_other_user_sudo(str(self), user_name)
+
```
katlapinka: I think it can be removed

Ostrokrzew: Removed.

connection/base_executor.py
```diff
@@ -43,6 +43,14 @@ def run_in_background(self, command):
...
             return int(output.stdout)
 
+    def run_as_other_user(self, command, user_name: str, sudo: bool = False):
+        prefix = f'sudo -u {user_name}'
+        if sudo:
+            command = 'sudo ' + command
+        command = f'{prefix} "{command}"'
+        output = self.run(command)
+        return output
+
```
robertbaldyga: IMO this function only obfuscates the code. It would be much more readable if the `sudo` were just included into the command in the test case code.
Moreover it's use case is so narrow, that it definitely should not be part of the framework.

Ostrokrzew: So I'll move it to my test avoid placing useless methods like this in test framework.

PR #92, opened 2019-12-17
Raise exception when failed to create filesystem or mount device

katlapinka:


test_tools/disk_utils.py
```diff
@@ -45,17 +45,13 @@ def create_filesystem(device, filesystem: Filesystem, force=True, blocksize=None
...
-    if output.exit_code == 0:
-        TestRun.LOGGER.info(
-            f"Successfully created filesystem on device: {device.system_path}")
-        return True
-
-    TestRun.LOGGER.error(
-        f"Could not create filesystem: {output.stderr}\n{output.stdout}")
-    return False
+    if output.exit_code != 0:
+        raise Exception(f"Could not create filesystem: {output.stderr}\n{output.stdout}")
```
robertbaldyga: Please use `CmdException` instead.

katlapinka: Done

PR #94, opened 2019-12-17
Add reboot functionality

robertbaldyga:


connection/ssh_executor.py
```diff
@@ -76,6 +76,13 @@ def rsync(self, src, dst, delete=False, symlinks=False, exclude_list=[],
     def is_remote(self):
         return True
 
+    def reboot(self):
+        try:
+            self.run("shutdown --reboot now")
+        except Exception:
+            pass
```
katlapinka: I think there should be raised some kind of exception, so we can handle it elsewhere and for example make hard reset when shutdown command could not be executed

connection/ssh_executor.py
```diff
@@ -76,6 +76,13 @@ def rsync(self, src, dst, delete=False, symlinks=False, exclude_list=[],
...
+        self.wait_for_connection()
```
katlapinka: I think there should be raised some kind of exception, so we can handle it elsewhere and for example make hard reset when shutdown command could not be executed

Deixx: If the shutdown for some reason fails, self.wait_for_connection() will be able to connect to DUT and will not wait at all.
I recommend adding a method that waits for loss of connection. Then wait for the connection to get up.

Deixx: In below scenario DUT did reboot, but the test performed following steps before the DUT went down.
2019/12/18 15:25:44 INFO:	Reboot
2019/12/18 15:25:44 INFO:	Check if cache is started
2019/12/18 15:25:45 INFO:	Check if core is added
2019/12/18 15:25:45 INFO:	Check if filesystem is mounted
2019/12/18 15:25:45 INFO:	Check if md5 checksum matches
2019/12/18 15:25:46 INFO:	Test cleanup
2019/12/18 15:25:46 INFO:	Unmounting device /dev/cas1-1.
2019/12/18 15:25:46 INFO:	Stop all caches
PASSED

robertbaldyga: Fixed.

PR #97, opened 2019-12-24
API for blktrace tool

Deixx:
+ utils

### DIFF
--- a/test_tools/blktrace.py
+++ b/test_tools/blktrace.py
@@ -1,0 +1,177 @@
+#
+# Copyright(c) 2019 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+import math
+
+from aenum import IntFlag, Enum
+
+from core.test_run import TestRun
+from storage_devices.device import Device
+from test_utils.filesystem.directory import Directory
+from test_utils.os_utils import is_mounted, drop_caches, DropCachesMode
+from test_utils.size import Size, Unit
+
+DEBUGFS_MOUNT_POINT = "/sys/kernel/debug"
+PREFIX = "trace_"
+HEADER_FORMAT = "%a|%C|%d|%e|%n|%N|%S|%5T.%9t\\n"
+
+
+class BlkTraceMask(IntFlag):
+    read = 1
+    write = 1 << 1
+    flush = 1 << 2
+    sync = 1 << 3
+    queue = 1 << 4
+    requeue = 1 << 5
+    issue = 1 << 6
+    complete = 1 << 7
+    fs = 1 << 8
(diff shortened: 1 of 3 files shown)


test_tools/blktrace.py
```diff
@@ -0,0 +1,177 @@
...
+    def start_monitoring(self, buffer_size: Size = None, number_of_subbuffers: int = None):
+        if self.blktrace_pid != -1:
+            raise Exception(f"blktrace already running with PID: {self.blktrace_pid}")
+
+        self.__outputDirectoryPath = Directory.create_temp_directory().full_path
+
+        drop_caches(DropCachesMode.ALL)
+
+        number_of_subbuffers = ("" if number_of_subbuffers is None
+                                else f" --num-sub-buffers = {number_of_subbuffers}")
```
Ostrokrzew: In other command there are no spaces between param and value.

test_utils/os_utils.py
```diff
@@ -70,6 +71,13 @@ def is_kernel_module_loaded(module_name):
     return output.exit_code == 0
 
 
+def is_mounted(path: str):
```
Ostrokrzew: In test framework in 'Device' class there's a method named like this. This shouldn't cause any conflicts, but You should check it.

test_tools/blktrace.py
Deixx: Done

test_tools/blktrace.py
```diff
@@ -0,0 +1,177 @@
...
+    F = 1 << 3  # Flush
+    S = 1 << 4  # Synchronous
+    N = 1 << 5  # None of the above
+
+
+class BlkTrace:
+    def __init__(self, device: Device, masks: BlkTraceMask = None):
+        self._mount_debugfs()
+        if device is None:
+            raise Exception("Wrong device provided (None)")
```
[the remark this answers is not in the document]
Deixx: :D

test_tools/blktrace.py
```diff
@@ -0,0 +1,177 @@
...
+        self.device = device
+        self.masks = "" if masks is None else f' -a {" -a ".join([m.name for m in masks])}'
+        self.blktrace_pid = -1
+        self.__outputDirectoryPath = None
+
+    @staticmethod
```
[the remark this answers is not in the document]
Deixx: It's not using anything BlkTrace class related, but it's within the class as it's not really useful elsewhere

test_utils/os_utils.py
Deixx: Device.is_mounted() checks if given instance of Device is mounted - I needed more general function that would check if given path (in this case a mountpoint) is used in mounting

PR #102, opened 2020-01-03
Add kedr wrappers 

micrakow:


test_tools/kedr.py
```diff
@@ -0,0 +1,118 @@
...
+    MEM_LEAK_CHECK = "leak_check.conf"
+    FAULT_SIM = "fsim.conf"
+
+
+class Kedr:
+    @staticmethod
+    def is_installed():
+        return "KEDR version" in TestRun.executor.run("kedr --version").stdout.strip()
+
+    def mount_debugfs():
```
Deixx: make it static or add 'self' as argument

test_tools/kedr.py
```diff
@@ -0,0 +1,118 @@
...
+        # TODO check if cmake is installed before
+        # TODO consider using os_utils.download_file()
+        kedr_archive = wget.download(kedr_0_6_url)
+
+        TestRun.executor.rsync(
+            kedr_archive,
+            f"{TestRun.plugins['opencas'].working_dir}")
+
+        kedr_dir = TestRun.executor.run(
+            f"cd {TestRun.plugins['opencas'].working_dir} && "
```
Deixx: Robert is changing plugins['opencas'] in https://github.com/Open-CAS/open-cas-linux/pull/263 - be aware of that
Also - I'm not sure if that's to be used in test-framework

test_tools/kedr.py
micrakow: Moved mount_debugfs() to os_utils

PR #103, opened 2020-01-03
Change 'create_filesystem' function in 'disk_utils' module

Ostrokrzew:
Wrapper for this checks if this function completes with success, but it isn't return anything.
Now this function changes 'filesystem' property of given object, not the wrapper. Before that change, 'filesystem' property didn't change in object of 'device' class (or subclasses).

Signed-off-by: Ostrokrzew <slawomir.jankowski@intel.com>

### DIFF
--- a/storage_devices/device.py
+++ b/storage_devices/device.py
@@ -17,8 +17,8 @@ class Device:
         self.mount_point = None
 
     def create_filesystem(self, fs_type: disk_utils.Filesystem, force=True, blocksize=None):
-        if disk_utils.create_filesystem(self, fs_type, force, blocksize):
-            self.filesystem = fs_type
+        disk_utils.create_filesystem(self, fs_type, force, blocksize)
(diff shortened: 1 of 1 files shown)


test_tools/disk_utils.py
```diff
@@ -53,6 +53,7 @@ def create_filesystem(device, filesystem: Filesystem, force=True, blocksize=None
         raise CmdException("Could not create filesystem.", output)
     TestRun.LOGGER.info(
         f"Successfully created filesystem on device: {device.system_path}")
+    device.filesystem = filesystem
```
micrakow: I'm not sure if method which is not a Device class member should change internal state of Device class object here, what do you think? How about leaving it how it was before, inside Device's create_filesystem method but just not check for return code at line :20?

Ostrokrzew: ok

PR #105, opened 2020-01-09
Catch undocumented exception from paramiko exec_command()

robertbaldyga:

### DIFF
--- a/connection/ssh_executor.py
+++ b/connection/ssh_executor.py
@@ -99,12 +99,12 @@ class SshExecutor(BaseExecutor):
                     continue
             raise ConnectionError("Timeout occurred while tying to establish ssh connection")
 
-    def wait_for_connection_loss(self, timeout: timedelta = timedelta(minutes=10)):
-        start_time = datetime.now()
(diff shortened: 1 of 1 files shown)


connection/ssh_executor.py
```diff
@@ -99,12 +99,12 @@ def wait_for_connection(self, timeout: timedelta = timedelta(minutes=10)):
                     continue
             raise ConnectionError("Timeout occurred while tying to establish ssh connection")
 
-    def wait_for_connection_loss(self, timeout: timedelta = timedelta(minutes=10)):
-        start_time = datetime.now()
+    def wait_for_connection_loss(self, timeout: timedelta = timedelta(minutes=1)):
```
Ostrokrzew: Should there be spaces "timedelta = timedelta"? I mean in other places, where You set default value to params, You don't place spaces between equal sign and words.
It's only aesthetic comment and You can ignore it.

robertbaldyga: I left spacing as it was before, but it looks like we have it like this in every place we use type annotation.

Issue #99, opened 2020-01-02
find_disks() does not correctly recognize OS disk if the OS is on an LVM partition.


tomaszrybicki:
https://github.com/Open-CAS/test-framework/blob/8b5d910c47b8e7c620bab1e551fbb6361c0d2448/test_utils/disk_finder.py#L131

The following disk configuration causes the OS disk to be added to the list of disks (which is not intended to have an OS disk):

    NAME                            MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
    sda                                      8:0        0   20G  0 disk 
    ├─sda1                                   8:1        0     1G  0 part   /boot
    └─sda2                                   8:2        0   19G  0 part 
      ├─fedora_localhost--live-root          253:0    0   17G  0 lvm    /

The command used for checking if the disk is OS disk ('mount | grep /') returns the following in this case:
``
/dev/mapper/fedora_localhost--live-root on / type ext4 (rw,relatime,seclabel)
``

PR #109, opened 2020-01-13
Disk finder fix

Deixx:
Fixes: https://github.com/Open-CAS/test-framework/issues/99

### DIFF
--- a/test_utils/disk_finder.py
+++ b/test_utils/disk_finder.py
@@ -5,6 +5,9 @@
 
 from core.test_run import TestRun
 from test_tools import disk_utils
+from test_tools.fs_utils import check_if_file_exists
+from test_utils import os_utils
+from test_utils.output import CmdException
 
 
 def find_disks():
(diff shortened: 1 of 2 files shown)


test_utils/disk_finder.py
```diff
@@ -127,6 +130,41 @@ def find_sata_ssd_device_path(serial_number, block_devices):
...
 
-def get_system_disk():
-    system_partition = TestRun.executor.run_expect_success('mount | grep " / "').stdout.split()[0]
-    return TestRun.executor.run_expect_success(f'lsblk -no pkname {system_partition}').stdout
+def get_system_disks():
+    system_device = TestRun.executor.run_expect_success('mount | grep " / "').stdout.split()[0]
+    readlink_output = \
+        TestRun.executor.run_expect_success(f'readlink -f {system_device}').stdout
+    device_name = readlink_output.split('/')[-1]
+    sys_block_path = os_utils.get_sys_block_path()
```
robertbaldyga: There is no reason to pass `sys_block_path` to `__get_slaves()` function as an argument. It would be much simpler if `__get_slaves()` called `os_utils.get_sys_block_path()` internally.

Deixx: Done

PR #110, opened 2020-01-14
Add 'wipe_filesystem' method and update 'FileProperties'

Ostrokrzew:

### DIFF
--- a/test_tools/disk_utils.py
+++ b/test_tools/disk_utils.py
@@ -299,3 +299,13 @@ def unit_to_string(unit):
         Unit.TeraByte: 'TB'
     }
     return unit_string.get(unit, "Invalid unit.")
+
+
+def wipe_filesystem(device, force=True):
+    TestRun.LOGGER.info(
+        f"Deleting filesystem ({device.filesystem.name}) on device: {device.system_path}")
+    force_param = ' -f' if force else ''
+    cmd = f'wipefs -a{force_param} {device.system_path}'
+    TestRun.executor.run_expect_success(cmd)
+    TestRun.LOGGER.info(
+        f"Successfully wiped filesystem from device: {device.system_path}")
--- a/test_utils/filesystem/file.py
+++ b/test_utils/filesystem/file.py
@@ -77,7 +77,6 @@ class FileProperties:
         self.group = file.group
         self.permissions = file.permissions
         self.size = file.size
-        self.md5 = file.md5
 
     @staticmethod
     def get_parent_dir(path):
@@ -94,6 +93,6 @@ class FileProperties:
         return tail or ntpath.basename(head)
 
     def __eq__(self, other):
-        return (self.md5 == other.md5 and self.permissions == other.permissions
-                and self.size == other.size and self.owner == other.owner
-                and self.group == other.group and self.name == other.name)
+        return (self.permissions == other.permissions and self.size == other.size
+                and self.owner == other.owner and self.group == other.group
+                and self.name == other.name)
--- a/storage_devices/device.py
+++ b/storage_devices/device.py
@@ -19,6 +19,10 @@ class Device:
         disk_utils.create_filesystem(self, fs_type, force, blocksize)
         self.filesystem = fs_type
 
+    def wipe_filesystem(self, force=True):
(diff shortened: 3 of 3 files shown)


test_tools/disk_utils.py
```diff
@@ -299,3 +299,17 @@ def unit_to_string(unit):
...
     }
     return unit_string.get(unit, "Invalid unit.")
+
+
+def wipe_filesystem(device, force=True):
+    TestRun.LOGGER.info(
+        f"Deleting filesystem ({device.filesystem.name}) on device: {device.system_path}")
+    force_param = '-f' if force else ''
+    cmd = f'wipefs -a {force_param} {device.system_path}'
+    cmd = re.sub(' +', ' ', cmd)
```
Deixx: Where did that ' +' come from? If either force_param or device.system_path contain ' +', then something is seriously wrong ;)

test_tools/disk_utils.py
```diff
@@ -299,3 +299,17 @@ def unit_to_string(unit):
         Unit.TeraByte: 'TB'
     }
     return unit_string.get(unit, "Invalid unit.")
+
+
+def wipe_filesystem(device, force=True):
+    TestRun.LOGGER.info(
+        f"Deleting filesystem ({device.filesystem.name}) on device: {device.system_path}")
+    force_param = '-f' if force else ''
+    cmd = f'wipefs -a {force_param} {device.system_path}'
```
Deixx: to avoid double space in command for force = False: change '-f' to ' -f' and remove space before {force_param}

test_utils/filesystem/file.py
```diff
@@ -14,6 +14,7 @@
 class File(FsItem):
     def __init__(self, full_path):
         FsItem.__init__(self, full_path)
+        self.md5 = self.md5sum()
```
Deixx: I don't think we want to count md5sum for every file created. And saving md5 to a file property is not a good idea. Small issue - contents of a file can be easily changed without updating its md5. Bigger issue - for large files calculating md5 can take quite a long time, therefore it should be calculated on demand.

robertbaldyga: We don't want it to be part of `File`. We have `FileProperties` for storing current file state.

Ostrokrzew: placed in `FileProperties`

Ostrokrzew: placed in `FileProperties`

test_tools/disk_utils.py
```diff
@@ -299,3 +299,16 @@ def unit_to_string(unit):
...
+
+
+def wipe_filesystem(device, force=True):
+    TestRun.LOGGER.info(
+        f"Deleting filesystem ({device.filesystem.name}) on device: {device.system_path}")
+    force_param = ' -f' if force else ''
+    cmd = f'wipefs -a{force_param} {device.system_path}'
+    output = TestRun.executor.run(cmd)
+    if output.exit_code != 0:
+        raise CmdException("Could not wipe filesystem.", output)
```
Deixx: can be replaced with TestRun.executor.run_expect_success(...)

test_tools/disk_utils.py
```diff
@@ -299,3 +299,16 @@ def unit_to_string(unit):
...
+    TestRun.LOGGER.info(
+        f"Successfully wiped filesystem from device: {device.system_path}")
+    return True
```
Deixx: remove (you never return False)

test_utils/filesystem/file.py
```diff
@@ -77,7 +77,7 @@ def __init__(self, file):
         self.group = file.group
         self.permissions = file.permissions
         self.size = file.size
-        self.md5 = file.md5
+        self.md5 = file.md5sum()
```
Deixx: I won't agree to non-explicit md5sum calculation. If someone wants md5 sum, they should call md5sum(), not get file properties. As I wrote before, calculating md5 sum can take a long time for large files, and should be done only on demand. The only solution I am willing to accept is saving md5sum in file when the md5sum() method is called.

Ostrokrzew: OK, let's do it Your way.

PR #114, opened 2020-01-22
Randomize test file

Ostrokrzew:
Change test file created by function to randomly generated string with custom lenght.
Remane function to more appropriate.

### DIFF
--- a/test_tools/fs_utils.py
+++ b/test_tools/fs_utils.py
@@ -5,12 +5,14 @@
 
 
 import base64
+import math
 import textwrap
 from datetime import datetime
 
 from aenum import IntFlag, Enum
 
 from core.test_run import TestRun
+from test_tools.dd import Dd
 from test_utils.size import Size, Unit
 
 
@@ -280,9 +282,17 @@ def parse_ls_output(ls_output, dir_path=''):
     return fs_items
 
 
-def create_test_file(target_file_path, file_content="Test content"):
+def create_random_test_file(target_file_path: str,
+                            file_size: Size = Size(1, Unit.MebiByte),
+                            random: bool = True):
     from test_utils.filesystem.file import File
+    bs = Size(512, Unit.KibiByte)
+    cnt = math.ceil(file_size.value / bs.value)
     file = File.create_file(target_file_path)
-    file.write(file_content)
+    dd = Dd().output(target_file_path) \
+             .input("/dev/urandom" if random else "/dev/zero") \
+             .block_size(bs) \
+             .count(cnt)
+    dd.run()
     file.refresh_item()
     return file
--- a/test_tools/fs_utils.py
+++ b/test_tools/fs_utils.py
@@ -67,9 +67,9 @@ def copy(source: str,
          recursive: bool = False,
          dereference: bool = False):
     cmd = f"cp{' --force' if force else ''}" \
-        f"{' --recursive' if recursive else ''}" \
-        f"{' --dereference' if dereference else ''} " \
-        f"{source} {destination}"
+          f"{' --recursive' if recursive else ''}" \
+          f"{' --dereference' if dereference else ''} " \
+          f"{source} {destination}"
     return TestRun.executor.run_expect_success(cmd)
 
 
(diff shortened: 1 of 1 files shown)


test_tools/fs_utils.py
```diff
@@ -280,9 +282,11 @@ def parse_ls_output(ls_output, dir_path=''):
...
 
 
-def create_test_file(target_file_path, file_content="Test content"):
+def create_random_test_file(target_file_path: str,
+                            file_size: Size = Size(1, Unit.MebiByte)):
     from test_utils.filesystem.file import File
     file = File.create_file(target_file_path)
-    file.write(file_content)
+    file.write("".join(random.choice(string.digits + string.ascii_letters + string.punctuation)
+                       for _ in range(file_size.value)))
```
robertbaldyga: `file.write()` transfers data from controller to DUT via ssh and it should never be used to write files bigger that few megabytes. Some test cases may want to create files that are much bigger, thus it would be better to have method that generates file entirely on DUTs side.

test_tools/fs_utils.py
```diff
@@ -65,9 +67,9 @@ def copy(source: str,
          recursive: bool = False,
          dereference: bool = False):
     cmd = f"cp{' --force' if force else ''}" \
-        f"{' --recursive' if recursive else ''}" \
-        f"{' --dereference' if dereference else ''} " \
-        f"{source} {destination}"
+          f"{' --recursive' if recursive else ''}" \
+          f"{' --dereference' if dereference else ''} " \
+          f"{source} {destination}"
     return TestRun.executor.run_expect_success(cmd)
```
robertbaldyga: Coding style improvements should be separated into another patch.

test_tools/fs_utils.py
Ostrokrzew: I modified it to use DD. Is it better solution?

robertbaldyga: It's better but not perfect. My biggest concern is `block_size(Size(1, Unit.Byte))`. 
Consider following results:
```
# dd if=/dev/zero of=/dev/nvme0n1 bs=131072 count=800
800+0 records in
800+0 records out 
104857600 bytes (105 MB, 100 MiB) copied, 0.136058 s, 771 MB/s

# dd if=/dev/zero of=/dev/nvme0n1 bs=4096 count=25600
25600+0 records in
25600+0 records out 
104857600 bytes (105 MB, 100 MiB) copied, 0.175051 s, 599 MB/s

# dd if=/dev/zero of=/dev/nvme0n1 bs=512 count=204800
204800+0 records in
204800+0 records out 
104857600 bytes (105 MB, 100 MiB) copied, 0.851134 s, 123 MB/s

# dd if=/dev/zero of=/dev/nvme0n1 bs=1 count=104857600
104857600+0 records in
104857600+0 records out 
104857600 bytes (105 MB, 100 MiB) copied, 136.207 s, 770 kB/s
```
In each case we write 100MiB with different block sizes. Difference between fastest and slowest case is ~100000%.  We **really** want to have bs=512 or bigger.

Ostrokrzew: Gave 1 MiB

Ostrokrzew: Gave 1 MiB.

test_tools/fs_utils.py
```diff
@@ -280,9 +281,16 @@ def parse_ls_output(ls_output, dir_path=''):
...
+                            file_size: Size = Size(1, Unit.MebiByte),
+                            random: bool = True):
     from test_utils.filesystem.file import File
+    bs = Size(1, Unit.MebiByte)
     file = File.create_file(target_file_path)
-    file.write(file_content)
+    dd = Dd().output(target_file_path) \
+             .input("/dev/urandom" if random else "/dev/zero") \
+             .block_size(bs) \
+             .count(int(file_size.value / bs.value))
```
robertbaldyga: So now if we try to create file smaller than 1MiB, the count will be 0.

Ostrokrzew: Changed to 512 KiB and add check for value under 1

test_tools/fs_utils.py
```diff
@@ -280,9 +281,17 @@ def parse_ls_output(ls_output, dir_path=''):
     return fs_items
 
 
-def create_test_file(target_file_path, file_content="Test content"):
+def create_random_test_file(target_file_path: str,
+                            file_size: Size = Size(1, Unit.MebiByte),
+                            random: bool = True):
     from test_utils.filesystem.file import File
+    bs = Size(512, Unit.KibiByte)
+    cnt = int(file_size.value / bs.value)
```
robertbaldyga: It would be better to have entire `cnt` calculation procedure in one place:
```python
cnt = math.ceil(file_size.value / bs.value)
```

PR #116, opened 2020-01-27
Added getting device's filesystem

Deixx:

### DIFF
--- a/storage_devices/device.py
+++ b/storage_devices/device.py
@@ -5,6 +5,7 @@
 
 from core.test_run import TestRun
 from test_tools import disk_utils
+from test_tools.disk_utils import get_device_filesystem_type
 from test_utils.io_stats import IoStats
 from test_utils.size import Size, Unit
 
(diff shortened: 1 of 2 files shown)


storage_devices/device.py
```diff
@@ -13,7 +14,7 @@ class Device:
     def __init__(self, path):
         self.size = Size(disk_utils.get_size(path.replace('/dev/', '')), Unit.Byte)
         self.system_path = path
-        self.filesystem = None
+        self.filesystem = get_device_filesystem_type(path)
```
Ostrokrzew: Debugger says that `path` is "/dev/XXX" but in this function You need only the end of path - this XXX. Change argument to `path.replace('/dev/', '')`
or to `path.split('/')[2]`
or modify `cmd = f'lsblk -l -o NAME,FSTYPE | grep "{device_system_path} "'` in called method similar way.

test_tools/disk_utils.py
```diff
@@ -309,3 +309,17 @@ def wipe_filesystem(device, force=True):
     TestRun.executor.run_expect_success(cmd)
     TestRun.LOGGER.info(
         f"Successfully wiped filesystem from device: {device.system_path}")
+
+
+def get_device_filesystem_type(device_system_path):
+    cmd = f'lsblk -l -o NAME,FSTYPE | grep "{device_system_path} "'
```
Ostrokrzew: whitespace

storage_devices/device.py
Deixx: Done

test_tools/disk_utils.py
Deixx: this whitespace is intentional to eliminate partitions from output if a device with partitions is checked

PR #117, opened 2020-01-29
Modify run_in_backgroud method

### LINKED ISSUE #115: If in config variable `env` is set, method 'run_in_background' doesn't work properly with casadm's commands.
In title I mean `env: "HTTPS_PROXY=myproxy.example:12345"` (example_config_file).
I have checked it with core flush, cache flush, cache start, core remove and cache mode switch commands when parameters were set for this operations to take some time.
It doesn't work with because of prefix (this `env` variable) added in called 'run' method:
`command = f"{TestRun.dut.env} && {command}"`.
This variable concatenated at start of the real command spoils ___casadm___ operations called in background by PyTest.

Ostrokrzew:
Modify this method to work properly with casadm.
Closes #115 issue

### DIFF
--- a/connection/base_executor.py
+++ b/connection/base_executor.py
@@ -28,8 +28,6 @@ class BaseExecutor:
         pass
 
     def run(self, command, timeout: timedelta = timedelta(minutes=30)):
-        if TestRun.dut and TestRun.dut.env:
-            command = f"{TestRun.dut.env} && {command}"
         command_id = TestRun.LOGGER.get_new_command_id()
         TestRun.LOGGER.write_command_to_command_log(command, command_id)
         output = self._execute(command, timeout)
@@ -37,7 +35,7 @@ class BaseExecutor:
         return output
 
     def run_in_background(self, command):
-        command += "&> /dev/null &echo $!"
+        command += " &> /dev/null &echo $!"
         output = self.run(command)
 
         if output is not None:


connection/base_executor.py
```diff
@@ -36,9 +36,12 @@ def run(self, command, timeout: timedelta = timedelta(minutes=30)):
...
 
-    def run_in_background(self, command):
-        command += "&> /dev/null &echo $!"
-        output = self.run(command)
+    def run_in_background(self, command, timeout: timedelta = timedelta(minutes=5)):
+        command += " &> /dev/null &echo $!"
+        command_id = TestRun.LOGGER.get_new_command_id()
+        TestRun.LOGGER.write_command_to_command_log(command, command_id)
+        output = self._execute(command, timeout)
+        TestRun.LOGGER.write_output_to_command_log(output, command_id)
```
katlapinka: Those lines are exactly the same as these in self.run() method, so I don't see any advantages of this change.

Ostrokrzew: Maybe when it looks like this would satisfy You. Advantage of this is that it works. Now it looks more like before, modification is smaller and more aesthetic. Added default flag to main 'run' method, which is swapped by 'run_in_background' method.

Ostrokrzew: Advantage of this is that it works. Now it looks more like before, modification is smaller and more aesthetic. Added default flag to main 'run' method, which is swapped by 'run_in_background' method.

connection/base_executor.py
```diff
@@ -27,8 +27,8 @@ def is_active(self):
     def wait_for_connection(self):
         pass
 
-    def run(self, command, timeout: timedelta = timedelta(minutes=30)):
-        if TestRun.dut and TestRun.dut.env:
+    def run(self, command, timeout: timedelta = timedelta(minutes=30), bckgd: bool = False):
```
Deixx: 2 things:
if you look at this method without looking at the next method then bckgd means nothing and there is no gain in this short version
I'd rather have an explanation why the old version doesn't work than implement some solution which works without investigating

Ostrokrzew: It doesn't work with casadm because of that prefix added in next line:
`command = f"{TestRun.dut.env} && {command}"`.
Without this it works good. This proxy command concatenated at start of the real command spoils casadm operations called in background.
Investigation made me think that this prefix is completely unnecessary. I tried few tests without this line and they stills works normally. I suggest removing this ___if___.
Please, tell me, if I'm wrong.

Deixx: What you wrote is not an explanation - it's problem statement.
You need to dig deeper - why doesn't it work with the prefix? Why would the prefix affect the execution of the command? I see no reason why it should.

PR #119, opened 2020-01-31
Extend 'Size' class methods

Ostrokrzew:

### DIFF
--- a/test_utils/size.py
+++ b/test_utils/size.py
@@ -75,23 +75,77 @@ class Size:
     def __add__(self, other):
         return Size(self.get_value() + other.get_value())
 
+    @multimethod
     def __lt__(self, other):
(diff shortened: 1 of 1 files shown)


test_utils/size.py
```diff
@@ -75,24 +75,78 @@ def __int__(self):
...
 
+    @multimethod
+    def __lt__(self, other: int):
+        return self.get_value() < other
+
+    @multimethod
+    def __lt__(self, other: float):
+        return self.get_value() < other
+
+    @multimethod
```
Deixx: veto - comparing size vs int gets a great NOPE from me

Ostrokrzew: I wanted this in case when some stats returns sizes and I want to compare them to zero, but OK, this is no problem. I would this close PR.

PR #121, opened 2020-02-04
Add iostat module

Ostrokrzew:
Needed to check IO statistics regardless of cache statistics, e.g. before cache start or after cache stop.

### DIFF
--- a/test_tools/iostat.py
+++ b/test_tools/iostat.py
@@ -1,0 +1,57 @@
+#
+# Copyright(c) 2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+from core.test_run import TestRun
+from storage_devices.device import Device
+from test_utils.size import Size, Unit
+
+
+class IOstat:
+    def __init__(self, stat_list: list = None):
+        if stat_list is None:
+            stat_list = [0, 0, 0, 0, 0]
+        if len(stat_list) < 5:
+            stat_list.extend([0, 0, 0, 0, 0])
+        self.transfers_per_second = float(stat_list[0])
+        self.reads_per_second = Size(float(stat_list[1]), Unit.KibiByte)
+        self.writes_per_second = Size(float(stat_list[2]), Unit.KibiByte)
+        self.total_reads = Size(float(stat_list[3]), Unit.KibiByte)
+        self.total_writes = Size(float(stat_list[4]), Unit.KibiByte)
+
+    def __str__(self):
+        return (
+            f"IO stats:\n"
+            f"Transfers per second: {self.transfers_per_second}\n"
+            f"Kilobytes read per second: {self.reads_per_second}\n"
+            f"Kilobytes written per second: {self.writes_per_second}\n"
+            f"Kilobytes read: {self.total_reads}\n"
+            f"Kilobytes written: {self.total_writes}\n"
+        )
+
+    def __eq__(self, other):
+        if not other:
+            return False
+        return (
+            self.transfers_per_second == other.transfers_per_second
+            and self.reads_per_second == other.reads_per_second
+            and self.writes_per_second == other.writes_per_second
+            and self.total_reads == other.total_reads
+            and self.total_writes == other.total_writes
+        )
+
+    @staticmethod
+    def get_iostat(device: Device):
+        """Returns object of IOstat class containing basic statistics displayed
+           in kibibytes/kibibytes per second."""
+        lines = TestRun.executor.run(
+            f"iostat -dk {device.system_path} "
+            f"| grep {device.system_path.strip('/dev/')}").stdout.splitlines()
+        iostat_list = []
+        for line in lines:
+            args = line.split()
+            if args[0] == device.system_path.strip('/dev/'):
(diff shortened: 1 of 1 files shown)


test_tools/iostat.py
```diff
@@ -0,0 +1,34 @@
...
+class IostatStatsEnum(IntEnum):
+    tps = 2,
+    blk_rdps = 3,
+    blk_wrtps = 4,
+    blk_reads = 5,
+    blk_writes = 6
+
+
+def get_iostat(device: Device, stat_filter: IostatStatsEnum = None):
+    """Returns one of basic statistics displayed in kB/ kBps as a float number."""
```
Deixx: remove space after /

test_tools/iostat.py
```diff
@@ -0,0 +1,34 @@
...
+    output = TestRun.executor.run(f"iostat -dk {device.system_path}"
+                                  f" | grep {device.system_path.replace('/dev/', '')}"
+                                  f" | awk 'NR==1 {{print ${stat_filter.value}}}'").stdout
+    return float(output)
+
+
+def get_iostat_all(device: Device):
+    """Returns list of basic statistics displayed in kB/ kBps."""
```
Deixx: the space

test_tools/iostat.py
```diff
@@ -0,0 +1,34 @@
...
+from core.test_run import TestRun
+from storage_devices.device import Device
+
+
+class IostatStatsEnum(IntEnum):
+    tps = 2,
+    blk_rdps = 3,
+    blk_wrtps = 4,
+    blk_reads = 5,
+    blk_writes = 6
```
Deixx: why blk if those are really kB?

test_tools/iostat.py
```diff
@@ -0,0 +1,34 @@
...
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+from aenum import IntEnum
+
+from core.test_run import TestRun
+from storage_devices.device import Device
+
+
+class IostatStatsEnum(IntEnum):
```
Deixx: we know it's an Enum - what does it enumerate? BUT FIRST! - read my comment to line 25

test_tools/iostat.py
```diff
@@ -0,0 +1,34 @@
...
+    blk_reads = 5,
+    blk_writes = 6
+
+
+def get_iostat(device: Device, stat_filter: IostatStatsEnum = None):
+    """Returns one of basic statistics displayed in kB/ kBps as a float number."""
+    output = TestRun.executor.run(f"iostat -dk {device.system_path}"
+                                  f" | grep {device.system_path.replace('/dev/', '')}"
+                                  f" | awk 'NR==1 {{print ${stat_filter.value}}}'").stdout
+    return float(output)
```
Deixx: I think methods here should return IoStats class which would have the output parsed to float properties. So you can call for example:
`get_iostat(device).tps`
this way we probably would need just one method

test_tools/iostat.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def get_iostat_all(device: Device):
```
Deixx: Why would you return float from one method and string[] from the other?

test_tools/iostat.py
```diff
@@ -0,0 +1,55 @@
...
+from core.test_run import TestRun
+from storage_devices.device import Device
+
+
+class IOstat:
+    def __init__(self, stat_list: list = None):
+        if stat_list is None or len(stat_list) < 5:
+            stat_list.extend([0, 0, 0, 0, 0])
+        if len(stat_list) > 5:
+            del stat_list[5:]
```
robertbaldyga: But actually why? If we are interested only in first 5 elements, then let's just ignore rest of them. I see no reason to delete elements from the list.

Ostrokrzew: okey

test_tools/iostat.py
```diff
@@ -0,0 +1,53 @@
...
+#
+
+from core.test_run import TestRun
+from storage_devices.device import Device
+
+
+class IOstat:
+    def __init__(self, stat_list: list = None):
+        if stat_list is None or len(stat_list) < 5:
+            stat_list.extend([0, 0, 0, 0, 0])
```
robertbaldyga: `AttributeError: 'NoneType' object has no attribute 'extend'`
If you really need to handle `None` here, you need to do it another way.

Ostrokrzew: thanks for warning

test_tools/iostat.py
```diff
@@ -0,0 +1,55 @@
...
+    def __init__(self, stat_list: list = None):
+        if stat_list is None:
+            stat_list = [0, 0, 0, 0, 0]
+        if len(stat_list) < 5:
+            stat_list.extend([0, 0, 0, 0, 0])
+        self.tps = float(stat_list[0])
+        self.read_kBps = float(stat_list[1])
+        self.write_kBps = float(stat_list[2])
+        self.kB_reads = float(stat_list[3])
+        self.kB_writes = float(stat_list[4])
```
mmichal10: Instead of raw floats, you can parse it to Size()

PR #122, opened 2020-02-06
Blktrace requeue action

mmichal10:


test_tools/blktrace.py
```diff
@@ -175,3 +176,23 @@ def parse(header_line: str):
...
+        if self.block_count:
+            ret += f"block_count: {self.block_count} "
+        if self.byte_count:
+            ret += f"byte_count: {self.byte_count} "
+        if self.command:
+            ret += f"command: {self.command} "
+        if self.error_value:
+            ret += f"error_value: {self.error_value} "
+        if self.rwbs:
+            ret += f"rwbs: {self.rwbs} "
```
katlapinka: maybe add __str__ method to rwbs because current name and value could be hard to read and understand

test_tools/blktrace.py
```diff
@@ -175,3 +176,23 @@ def parse(header_line: str):
         header.timestamp = int(timestamp_fields[0]) * math.pow(10, 9) + timestamp_nano
 
         return header
+
+    def __str__(self):
+        ret = ""
+        if self.action:
+            ret += f"action: {self.action} "
```
katlapinka: .name
and maybe in brackets ('{self.action.value}')

PR #131, opened 2020-02-27
Custom class for time representation.

mmichal10:
Builtin python types don't have capabilities to retrieve nanoseconds.

### DIFF
--- a/test_utils/time.py
+++ b/test_utils/time.py
@@ -1,0 +1,11 @@
+#
+# Copyright(c) 2019-2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+from attotime import attotimedelta
+
+
+class Time(attotimedelta):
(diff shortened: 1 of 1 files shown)


test_utils/time.py
```diff
@@ -0,0 +1,11 @@
...
+# Copyright(c) 2019-2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+from attotime import attotimedelta
+
+
+class Time(attotimedelta):
+    def total_miliseconds(self):
+        self.total_nanoseconds() / 1000
```
Ostrokrzew: Maybe this would be better:
```python
def total_miliseconds(self):
     return self.total_nanoseconds() / constants.NANOSECONDS_PER_MILLISECOND
```
_NANOSECONDS_PER_MILLISECOND equals 1 000 000

PR #133, opened 2020-03-03
Allow to use all possible fio outputs

micrakow:


test_tools/fio/fio.py
```diff
@@ -21,13 +21,13 @@ def __init__(self, executor_obj=None):
         self.base_cmd_parameters: test_tools.fio.fio_param.FioParam = None
         self.global_cmd_parameters: test_tools.fio.fio_param.FioParam = None
 
-    def create_command(self):
+    def create_command(self, output_type=test_tools.fio.fio_param.FioOutput.json):
```
Ostrokrzew: There is `import test_tools.fio.fio_param` in this file so this parameter could be shorter.

PR #136, opened 2020-03-04
Fix gathering of 'messages' log

rafalste:
On Linux systems using journald instead of syslog (e.g. Ubuntu) there's
no /var/log/messages log, so we need to gather /var/log/syslog instead,
which also contains everything that used to be in /var/log/messages.

### DIFF
--- a/log/logger.py
+++ b/log/logger.py
@@ -187,9 +187,17 @@ class Log(HtmlLogManager, metaclass=Singleton):
                 log_destination_path = os.path.join(self.base_dir, "dut_info", f'{key}.log')
                 if type(TestRun.executor) is SshExecutor:
                     sftp = TestRun.executor.ssh.open_sftp()
+                    if key is "messages":
+                        try:
+                            sftp.stat(log_files[key])
+                        except FileNotFoundError:
+                            log_files[key] = "/var/log/syslog"
                     sftp.get(log_files[key], log_destination_path)
                     sftp.close()
                 elif type(TestRun.executor) is LocalExecutor:
+                    if key is "messages":
+                        if not os.path.isfile(log_files[key]):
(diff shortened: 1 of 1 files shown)


log/logger.py
```diff
@@ -177,7 +177,9 @@ def get_additional_logs(self):
         from core.test_run import TestRun
         from connection.ssh_executor import SshExecutor
         from connection.local_executor import LocalExecutor
-        log_files = {"messages": "/var/log/messages",
+        messages_log = "/var/log/messages"
+        sys_log = "/var/log/syslog"
```
robertbaldyga: Change it to:
```python
messages_log = "/var/log/messages"
if not check_if_file_exists(messages_log):
    messages_log = "/var/log/syslog"
```
And the rest of this patch will be unneeded.

log/logger.py
```diff
@@ -187,9 +189,17 @@ def get_additional_logs(self):
                 log_destination_path = os.path.join(self.base_dir, "dut_info", f'{key}.log')
                 if type(TestRun.executor) is SshExecutor:
```
robertbaldyga: Huh, I've just noticed these checks by executor type and they clearly violate Executor abstraction.

log/logger.py
rafalste: Actually we need to check this log existence in two scenarios - local and remote test execution.
- check if 'sftp.stat(file)' fails if the test is executed remotely
- or check 'os.path.isfile(file)' if test is executed locally

robertbaldyga: And `check_if_file_exists()` function does exactly that - it checks if file exists in a way specific for given executor. That way we have those two scenarios abstracted out and we can write simple generic code.

PR #138, opened 2020-03-07
Add bidirectional rsync methods

rafalste:

### DIFF
--- a/connection/base_executor.py
+++ b/connection/base_executor.py
@@ -15,8 +15,16 @@ class BaseExecutor:
     def _execute(self, command, timeout):
         raise NotImplementedError()
 
-    def rsync(self, src, dst, delete, symlinks, exclude_list, timeout):
+    def _rsync(self, src, dst, delete, symlinks, exclude_list, timeout, dut_to_controller):
         raise NotImplementedError()
+
+    def rsync_to(self, src, dst, delete=False, symlinks=False, exclude_list=[],
+                 timeout: timedelta = timedelta(seconds=30)):
+        return self._rsync(src, dst, delete, symlinks, exclude_list, timeout, False)
+
+    def rsync_from(self, src, dst, delete=False, symlinks=False, exclude_list=[],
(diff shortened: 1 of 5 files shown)


connection/base_executor.py
```diff
@@ -18,6 +18,14 @@ def _execute(self, command, timeout):
...
         raise NotImplementedError()
 
+    def rsync_to(self, src, dst, delete=False, symlinks=False, exclude_list=[],
+                 timeout: timedelta = timedelta(seconds=30)):
+        return self.rsync(src, dst, delete, symlinks, exclude_list, timeout, False)
+
+    def rsync_from(self, src, dst, delete=False, symlinks=False, exclude_list=[],
+                   timeout: timedelta = timedelta(seconds=30)):
+        return self.rsync(src, dst, delete, symlinks, exclude_list, timeout, True)
+
```
Ostrokrzew: You added two exactly the same methods. For what purpose? Do You use them somewhere?

rafalste: Take a look at the last parameter of the method that is returned. In 'rsync_from' it's 'True' and in 'rsync_to' it's 'False'. Those booleans refer to 'dut_to_controller' argument of 'rsync' method which is responsible for direction of transfer (dut<=>controller).

connection/base_executor.py
```diff
@@ -18,6 +18,14 @@ def _execute(self, command, timeout):
     def rsync(self, src, dst, delete, symlinks, exclude_list, timeout):
         raise NotImplementedError()
 
+    def rsync_to(self, src, dst, delete=False, symlinks=False, exclude_list=[],
+                 timeout: timedelta = timedelta(seconds=30)):
+        return self.rsync(src, dst, delete, symlinks, exclude_list, timeout, False)
```
robertbaldyga: We may want to rename `rsync()` method to `_rsync()`, indicating that this method is private and `rsync_to()`/`rsync_from()` should be used instead. This will also require small changes in OCL repository.

PR #139, opened 2020-03-07
Fix gathering of 'messages' log

rafalste:
Dependent on https://github.com/Open-CAS/test-framework/pull/138

### DIFF
--- a/log/logger.py
+++ b/log/logger.py
@@ -177,20 +177,19 @@ class Log(HtmlLogManager, metaclass=Singleton):
         from core.test_run import TestRun
         from connection.ssh_executor import SshExecutor
         from connection.local_executor import LocalExecutor
-        log_files = {"messages": "/var/log/messages",
+        from test_tools.fs_utils import check_if_file_exists
+        messages_log = "/var/log/messages"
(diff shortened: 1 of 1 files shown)


log/logger.py
```diff
@@ -177,20 +177,19 @@ def get_additional_logs(self):
...
         for key in log_files.keys():
             try:
                 log_destination_path = os.path.join(self.base_dir, "dut_info", f'{key}.log')
-                if type(TestRun.executor) is SshExecutor:
-                    sftp = TestRun.executor.ssh.open_sftp()
-                    sftp.get(log_files[key], log_destination_path)
-                    sftp.close()
-                elif type(TestRun.executor) is LocalExecutor:
-                    TestRun.executor.run(f"cp {log_files[key]} {log_destination_path}")
+                TestRun.executor.rsync_from(log_files[key], log_destination_path)
```
imjfckm: Using `for key in log_files.keys()` and then referencing values by `log_files[key]`. Maybe change that to `for key, path in log_files.items()`?

rafalste: You're right, already fixed this!

PR #142, opened 2020-03-17
Add `attotime` package to requirements.txt

mmichal10:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,3 +9,4 @@ typing==3.7.4.1
 pyyaml>=5.1
 lxml>=4.4.1
 wget>=3.2
+attotime>=0.2.0


rafalste:
'attotime' is already there in requirements.txt file in open-cas-linux

requirements.txt
```diff
@@ -9,3 +9,4 @@ typing==3.7.4.1
 pyyaml>=5.1
 lxml>=4.4.1
 wget>=3.2
+attotime==0.2.0
```
Ostrokrzew: I approved this, but You should consider using there `greater or equal` sign in place of `equal` sign.

mmichal10: I fixed it and dismissed 4 approves😢

PR #144, opened 2020-03-30
Extends rsync by checksum option

mariuszbarczak:
Signed-off-by: Mariusz Barczak <mariusz.barczak@intel.com>

### DIFF
--- a/connection/ssh_executor.py
+++ b/connection/ssh_executor.py
@@ -51,7 +51,7 @@ class SshExecutor(BaseExecutor):
 
         return Output(stdout.read(), stderr.read(), stdout.channel.recv_exit_status())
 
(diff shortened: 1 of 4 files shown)


connection/ssh_executor.py
```diff
@@ -51,14 +51,16 @@ def _execute(self, command, timeout):
 
         return Output(stdout.read(), stderr.read(), stdout.channel.recv_exit_status())
 
-    def _rsync(self, src, dst, delete=False, symlinks=False, exclude_list=[],
+    def _rsync(self, src, dst, delete=False, symlinks=False, checksum=False, exclude_list=[],
```
robertbaldyga: 1. This is overloaded method. Please update it in base class and its all derivatives.
2. This method is private and I don't see this parameter used anywhere in the class. It may be good idea to expose this new functionality in public methods.

PR #145, opened 2020-04-02
Add new pytest marks

Ostrokrzew:


robertbaldyga:
Those should probably be moved to OCL repository. TF doesn't provide any functionalities related to kernel version, so TF user cannot benefit from these marks in any way.

PR #146, opened 2020-04-03
Fix unloading kernel modules

Ostrokrzew:
`run_expect_success` here causes test fails in setup phase,
when module wasn't loaded before.


test_utils/os_utils.py
```diff
@@ -81,7 +81,7 @@ def load_kernel_module(module_name, module_args: {str, str}=None):
 
 def unload_kernel_module(module_name, unload_method: ModuleRemoveMethod = ModuleRemoveMethod.rmmod):
     cmd = f"{unload_method.value} {module_name}"
-    return TestRun.executor.run_expect_success(cmd)
+    return TestRun.executor.run(cmd)
```
robertbaldyga, line 84: Throwing exception is preferred over returning object representing success status.

PR #151, opened 2020-04-30
Add get symlink target method

katlapinka:


test_utils/filesystem/symlink.py
```diff
@@ -18,3 +18,6 @@ def md5sum(self, binary=True):
             raise Exception(
                 f"Md5sum command execution failed! {output.stdout}\n{output.stderr}")
         return output.stdout.split()[0]
+
+    def refresh_target(self):
+        self.target = TestRun.executor.run(f"readlink -f {self.full_path}").stdout
```
robertbaldyga, line 23: Instead of having internal state that needs to be refreshed, maybe it would be better to introduce `get_target()` method, which would always return up-to-date result?

PR #154, opened 2020-05-08
Make sure that connection is lost on reboot

robertbaldyga:

### DIFF
--- a/connection/ssh_executor.py
+++ b/connection/ssh_executor.py
@@ -26,7 +26,11 @@ class SshExecutor(BaseExecutor):
     def __del__(self):
         self.ssh.close()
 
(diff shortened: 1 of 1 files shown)


connection/ssh_executor.py
```diff
@@ -101,6 +101,7 @@ def wait_for_connection(self, timeout: timedelta = timedelta(minutes=10)):
             while start_time + timeout > datetime.now():
                 try:
                     self.connect(user=self.user, passwd=self.password, port=self.port)
+                    self.ssh.exec_command(":", timeout=30)
```
katlapinka, line 104: why can't you use the method below - wait_for_connection after reboot?

robertbaldyga: Fixed.

PR #156, opened 2020-05-18
Log failed imports of optional plugins

mmichal10:
Missing plugins omitted without any notification lead to confusion.

### DIFF
--- a/core/plugins.py
+++ b/core/plugins.py
@@ -6,6 +6,7 @@
 import pytest
 import sys
 import importlib
+from core.test_run import TestRun
 
 
 class PluginManager:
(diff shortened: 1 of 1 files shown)


core/plugins.py
```diff
@@ -44,7 +46,8 @@ def __init__(self, item, config):
         for name, mod in opt_plugin_mod.items():
             try:
                 self.plugins[name] = mod.plugin_class(self.opt_plugins[name])
-            except Exception:
+            except Exception as e:
+                TestRun.LOGGER.debug(f"Failed to initialize {name} - optional plugin. Reason: {e}")
```
katlapinka, line 50: you can add ' ' arount {name} like in the debug message above

PR #157, opened 2020-05-19
Improve iostat wrapper

mmichal10:

### DIFF
--- a/test_tools/iostat.py
+++ b/test_tools/iostat.py
@@ -5,53 +5,118 @@
 
 from core.test_run import TestRun
 from storage_devices.device import Device
-from test_utils.size import Size, Unit
+from test_utils.size import Size, Unit, UnitPerSecond
+from test_utils.time import Time
 
 
 class IOstat:
-    def __init__(self, stat_list: list = None):
-        if stat_list is None:
-            stat_list = [0, 0, 0, 0, 0]
-        if len(stat_list) < 5:
-            stat_list.extend([0, 0, 0, 0, 0])
-        self.transfers_per_second = float(stat_list[0])
-        self.reads_per_second = Size(float(stat_list[1]), Unit.KibiByte)
-        self.writes_per_second = Size(float(stat_list[2]), Unit.KibiByte)
-        self.total_reads = Size(float(stat_list[3]), Unit.KibiByte)
-        self.total_writes = Size(float(stat_list[4]), Unit.KibiByte)
+    def __init__(self, device_name: str, raw_stats: list = None):
+        metrics_number = 13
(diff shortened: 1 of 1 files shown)


test_tools/iostat.py
```diff
@@ -5,53 +5,107 @@
...
-        self.total_writes = Size(float(stat_list[4]), Unit.KibiByte)
+    def __init__(self, device_name: str, raw_stats: list = None):
+        metrics_number = 13
+        if raw_stats is None:
+            raw_stats = [0] * metrics_number
+        if len(raw_stats) < metrics_number:
+            raw_stats.extend([0] * metrics_number)
+
+        self.device_name = device_name
+        self.read_requests_merged_per_second = float(raw_stats[0])
```
katlapinka, line 21: shouldn't it be Size, as write_requests_merged_per second below?

test_tools/iostat.py
```diff
@@ -5,53 +5,107 @@
...
+            f"Write requests: {self.write_requests}\n"
+            f"Reads per second: {self.reads_per_sec}\n"
+            f"Writes per second {self.writes_per_sec}\n"
+            f"Average request size: {self.average_request_size} blocks\n"
+            f"Average queue length {self.average_queue_length}\n"
+            f"Average service time: {self.average_service_time}\n"
+            f"Read average service time {self.read_average_service_time}\n"
+            f"Write average service time: {self.write_average_service_time}\n"
+            f"Utilization: {self.utilization}\n"
+            f"=================================================\n"
```
Ostrokrzew, line 50: If You want, You can change it to triple-quoted fstring.

test_tools/iostat.py
```diff
@@ -5,53 +5,107 @@
...
-            f"| grep {device.system_path.strip('/dev/')}").stdout.splitlines()
-        iostat_list = []
+            f"{iostat_cmd} " f'| grep "{grep_pattern}"'
+        ).stdout.splitlines()
+
+        ret = []
+        raw_stats = []
         for line in lines:
             args = line.split()
-            if args[0] == device.system_path.strip('/dev/'):
```
katlapinka, line 54: Why did you remove this if? What if the first element of line will be different than device name? e.g. some header in the first line

test_tools/iostat.py
mmichal10: Actually `write_requests_merged_per` should be raw float, but thanks for noticing it

test_tools/iostat.py
mmichal10: Since iostat output is filtered with `grep`, headers won't be included.

PR #158, opened 2020-05-19
Set reboot timeout in config

Ostrokrzew:
Make reboot timeout more customizable, because OpenCAS actually
has 30 min init timeout, so when dut initialize OpenCAS service longer
than 10 minutes (default timeout in `wait_for_connection` method),
`reboot` method fails and test using this method terminates early.
Now timeout could be passed in config file.

### DIFF
--- a/connection/ssh_executor.py
+++ b/connection/ssh_executor.py
@@ -23,6 +23,7 @@ class SshExecutor(BaseExecutor):
         self.port = port
         self.ssh = paramiko.SSHClient()
         self.connect(username, password, port)
+        self._check_config_for_reboot_timeout()
 
     def __del__(self):
         self.ssh.close()
@@ -88,10 +89,22 @@ class SshExecutor(BaseExecutor):
     def is_remote(self):
         return True
 
+    def _check_config_for_reboot_timeout(self):
+        if "reboot_timeout" in TestRun.config.keys():
+            self._parse_timeout_to_int()
+        else:
+            self.reboot_timeout = None
+
+    def _parse_timeout_to_int(self):
+        self.reboot_timeout = int(TestRun.config["reboot_timeout"])
+        if self.reboot_timeout < 0:
+            raise ValueError("Reboot timeout cannot be negative.")
+
     def reboot(self):
         self.run("reboot")
         self.wait_for_connection_loss()
-        self.wait_for_connection()
+        self.wait_for_connection(timedelta(seconds=self.reboot_timeout)) \
+            if self.reboot_timeout is not None else self.wait_for_connection()
 
     def is_active(self):
         try:


connection/ssh_executor.py
```diff
@@ -95,7 +95,7 @@ def is_active(self):
         except Exception:
             return False
 
-    def wait_for_connection(self, timeout: timedelta = timedelta(minutes=10)):
+    def wait_for_connection(self, timeout: timedelta = timedelta(minutes=40)):
```
robertbaldyga, line 98: If your use case is reboot, maybe it would be better to add `timeout` parameter to `reboot()` method?
BTW maybe timeout default for reboot should be settable in config? It's strongly correlated with platform (on bare metal reboot may take tens of minutes, while on VMs we rather expect tens of seconds) so having configurable parameter like that sounds reasonable.

core/test_run_utils.py
```diff
@@ -125,6 +125,15 @@ def __setup(cls):
...
 
+    if "reboot_timeout" in cls.config:
+        try:
+            cls.config["reboot_timeout"] = int(cls.config["reboot_timeout"])
+            if cls.config["reboot_timeout"] < 0:
+                cls.config["reboot_timeout"] = None
+        except Exception:
+            TestRun.LOGGER.info("Wrong reboot timeout value. Setting it to 'None'.")
+            cls.config["reboot_timeout"] = None
+
```
robertbaldyga, line 136: It should be moved to constructor of `ssh_executor`. That way reboot related logic will be encapsulated within `ssh_executor` class.

connection/ssh_executor.py
```diff
@@ -86,7 +86,8 @@ def is_remote(self):
     def reboot(self):
         self.run("reboot")
         self.wait_for_connection_loss()
-        self.wait_for_connection()
+        self.wait_for_connection(timedelta(minutes=TestRun.config["reboot_timeout"])) \
+            if TestRun.config["reboot_timeout"] is not None else self.wait_for_connection()
```
robertbaldyga, line 90: Timeout is typically expressed in seconds. Having timeout in minutes may be misleading.

connection/ssh_executor.py
```diff
@@ -22,6 +22,14 @@ def __init__(self, ip, username, password, port=22):
         self.port = port
         self.ssh = paramiko.SSHClient()
         self.connect(username, password, port)
+        # timeout for reboot
+        try:
+            self.reboot_timeout = int(TestRun.config["reboot_timeout"])
+            if self.reboot_timeout < 0:
+                self.reboot_timeout = None
```
robertbaldyga, line 29: Don't we want to throw exception in such case? It's good candidate for `ValueError` imo.

Ostrokrzew: You're right, done.

connection/ssh_executor.py
```diff
@@ -22,6 +22,14 @@ def __init__(self, ip, username, password, port=22):
...
         self.ssh = paramiko.SSHClient()
         self.connect(username, password, port)
+        # timeout for reboot
+        try:
+            self.reboot_timeout = int(TestRun.config["reboot_timeout"])
+            if self.reboot_timeout < 0:
+                raise ValueError
+        except Exception:
+            TestRun.LOGGER.info("Wrong reboot timeout value. Setting it to 'None'.")
+            self.reboot_timeout = None
```
robertbaldyga, line 32: There is no point in throwing exception and catching it few lines later. We want to stop executing test, due to invalid config option.

connection/ssh_executor.py
```diff
@@ -23,6 +23,14 @@ def __init__(self, ip, username, password, port=22):
         self.port = port
         self.ssh = paramiko.SSHClient()
         self.connect(username, password, port)
+        # timeout for reboot
+        try:
+            self.reboot_timeout = int(TestRun.config["reboot_timeout"])
```
katlapinka, line 28: maybe we can set some default timeout when there is no "reboot_timeout" specified in config?

Ostrokrzew: It actually has default value - 10 minutes - specified here:
`wait_for_connection(self, timeout: timedelta = timedelta(minutes=10))`

connection/ssh_executor.py
```diff
@@ -88,10 +89,20 @@ def _rsync(self, src, dst, delete=False, symlinks=False, checksum=False, exclude
     def is_remote(self):
         return True
 
+    def _check_config_for_reboot_timeout(self):
+        try:
+            self.reboot_timeout = int(TestRun.config["reboot_timeout"])
```
katlapinka, line 94: if there is no "reboot_timeout" in config, exception will be raised, because you catch exception in except block but a new exception is raised there - I suggest setting self.reboot_timeout as None when it's not specified in config

Ostrokrzew: Thanks, done.

PR #162, opened 2020-05-25
Add the IOstatBasic class

Ostrokrzew:
Create class named `IOstatBasic` and rename class: `IOstat` to `IOStatExtended`
Extract method `get_iostat_list`and add wrappers to both `IOstat` classes.
The `IOstatBasic` contain other stats than the `IOstatExtended` and is needed
in a few lazy-write tests.

### DIFF
--- a/test_tools/iostat.py
+++ b/test_tools/iostat.py
@@ -9,7 +9,9 @@ from test_utils.size import Size, Unit, UnitPerSecond
 from test_utils.time import Time
 
 
-class IOstat:
+class IOstatExtended:
+    iostat_option = "x"
+
     def __init__(self, device_name: str, raw_stats: list = None):
         metrics_number = 13
         if raw_stats is None:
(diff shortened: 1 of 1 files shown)


test_tools/iostat.py
```diff
@@ -10,14 +10,76 @@
...
+            args = line.split()
+
+            dev_name = args[0]
+            raw_stats = args[1:]
+            ret += [cls(device_name=dev_name, raw_stats=raw_stats)]
+
+        return ret
+
+
+class IOstatExtended(IOstat):
```
mmichal10, line 80: I would avoid using inheritance. You keep both `raw_stats` and fields with single stats which is redundant. `IOstatBasic` and `IOstatExtended` doesn't have much in common except they are both wrappers for `iostat` output.

test_tools/iostat.py
```diff
@@ -10,14 +10,76 @@
...
+    def __str__(self):
+        # must be overridden
+        return
+
+    def __repr__(self):
+        return str(self)
+
+    def __eq__(self, other):
+        if not other:
+            return False
```
mmichal10, line 36: Why is this needed? How would comparison would look like to fulfill it?

test_tools/iostat.py
```diff
@@ -84,39 +84,105 @@ def __eq__(self, other):
...
+            interval: int = 1,
+    ):
+        """
+            Returns list of IOstat objects containing basic statistics displayed
+            in kibibytes/kibibytes per second.
+        """
+        return get_iostat_list(cls, devices_list, iostat_cmd, since_boot, interval)
+
+
+def get_iostat_list(
```
mmichal10, line 156: If `get_iostat_list()` is 'private' fucntion - it won't be called from outside of this file you can rename it to `_get_iostat_list()` for better readability

test_tools/iostat.py
```diff
@@ -84,39 +84,105 @@ def __eq__(self, other):
...
+    def __eq__(self, other):
+        if not isinstance(other, IOstatBasic):
+            return False
+        return vars(self) == vars(other)
+
+    @classmethod
+    def get_iostat_list(
+            cls,
+            devices_list: [Device],
+            iostat_cmd: str = "iostat -dk ",
```
mmichal10, line 145: I think there is no need to allow caller pass whole iostat command so this parameter is redundant

test_tools/iostat.py
```diff
@@ -84,39 +84,105 @@ def __eq__(self, other):
             and self.utilization == other.utilization
         )
 
-    @staticmethod
+    @classmethod
     def get_iostat_list(
-        devices_list: [Device], since_boot: bool = True, interval: int = 1
+            cls,
+            devices_list: [Device],
+            iostat_cmd: str = "iostat -xk ",
```
mmichal10, line 91: I think there is no need to allow caller pass whole iostat command so this parameter is redundant. You can hardcode it inside this function

PR #163, opened 2020-05-29
Add some OS utilities to os_utils

Ostrokrzew:


test_utils/os_utils.py
```diff
@@ -159,3 +160,127 @@ def get_wbt_lat(device: Device):
...
+    default_grub_file = read_file(path_default_grub).splitlines()
+    grub_content = ""
+
+    for line in default_grub_file:
+        if "GRUB_DEFAULT" in line:
+            line = f"GRUB_DEFAULT={kernel_version}"
+        grub_content += line + "\n"
+
+    write_file(path_default_grub, grub_content)
+    TestRun.executor.run(f"{grub_version}-mkconfig -o /boot/{grub_version}/grub.cfg")
```
robertbaldyga, line 286: This entire logic, including transferring file back and forth between DUT and controller, could be replaced with single `sed -i` command.

Ostrokrzew: done

test_utils/os_utils.py
```diff
@@ -205,3 +206,121 @@ def get_wbt_lat(device: Device):
     )
 
     return int(TestRun.executor.run_expect_success(f"cat {wbt_lat_config_path}").stdout)
+
+
+def got_compatible_kernels():
+    count = 0
+    try:
+        count = _count_versions_in_kernels(compatible=True)
+
```
Deixx, line 215: don't put empty lines in single blocks please - just as if..else is a single block, so is try..except..finally

test_utils/os_utils.py
```diff
@@ -205,3 +206,121 @@ def get_wbt_lat(device: Device):
...
+    finally:
+        return True if count > 1 else False
+
+
+def got_incompatible_kernels():
+    kernels_number = 0
+    try:
+        kernels_number = _count_versions_in_kernels(compatible=False)
+
```
Deixx, line 228: same

test_utils/os_utils.py
```diff
@@ -205,3 +206,121 @@ def get_wbt_lat(device: Device):
...
+                count += 1
+
+    return count
+
+
+def _read_grub_version():
+    return TestRun.executor.run("ls /boot | grep grub").stdout.splitlines()[-1]
+
+
+def switch_kernel(compatible: bool = True, custom: str = None):
```
Deixx, line 261: What is your intention here?
Do you really need to change the default kernel in the OS or would calling 'grub2-reboot <kernel position in grub>' be enough? It sets kernel to boot to on next boot.
If you need to change the default, then 'grub2-set-default <kernel position in grub>' would do the trick.

test_utils/os_utils.py
```diff
@@ -381,3 +380,136 @@ def get_wbt_lat(device: Device):
...
+    try:
+        kernels_number = _count_versions_in_kernels(compatible=False)
+    except Exception as e:
+        raise Exception("Cannot check if there are incompatible kernels on DUT\n"
+                        f"{e}")
+    finally:
+        return True if kernels_number > 0 else False
+
+
+def _count_versions_in_kernels(compatible: bool):
```
Deixx, line 407: _count_kernel_versions would be a better name

test_utils/os_utils.py
```diff
@@ -381,3 +380,136 @@ def get_wbt_lat(device: Device):
...
+            if kernel_ver in kernel:
+                count += 1
+        else:
+            if kernel_ver not in kernel:
+                count += 1
+
+    return count
+
+
+def _read_grub_version():
```
Deixx, line 427: _get is kind of a standard in naming such methods ;)

test_utils/os_utils.py
```diff
@@ -381,3 +380,136 @@ def get_wbt_lat(device: Device):
...
+def choose_kernel_version(compatible: bool, available_kernels: [str], custom: str):
+    current_version = read_current_kernel()
+    current_version_main = str(get_kernel_version())
+
+    if custom:
+        for kernel in available_kernels:
+            kernel = kernel.split('--')[0]
+
+            if custom in kernel:
+                kernel_version = int(kernel.split("#")[0])
```
Deixx, line 451: From what I see this is not a kernel version - rather its number in grub config

test_utils/os_utils.py
```diff
@@ -381,3 +380,136 @@ def get_wbt_lat(device: Device):
...
+                break
+
+    if available_kernels and current_version:
```
Deixx, line 454: Let's say we got our "kernel_version"  from custom, so now we're going to overwrite that in this section

test_utils/os_utils.py
```diff
@@ -381,3 +380,136 @@ def get_wbt_lat(device: Device):
...
+        if compatible:
+            compatible_versions = _get_kernels(True, available_kernels, current_version_main)
+
+            for kernel in compatible_versions:
+                if current_version not in kernel:
+                    kernel_version = int(kernel.split("#")[0])
```
Deixx, line 460: add break

test_utils/os_utils.py
```diff
@@ -381,3 +380,136 @@ def get_wbt_lat(device: Device):
...
+        else:
+            incompatible_versions = _get_kernels(False, available_kernels, current_version_main)
+            kernel_version = int(incompatible_versions[0].split("#")[0])
+
+    return kernel_version
```
Deixx, line 465: not all paths return kernel_version - maybe add some exceptions (i.e. when custom version is not found)

test_utils/os_utils.py
```diff
@@ -381,3 +380,136 @@ def get_wbt_lat(device: Device):
...
+
+def _set_default_kernel(kernel_version: int, grub_version: str):
+    path_default_grub = "/etc/default/grub"
+    TestRun.executor.run(
+        f"sed -i '/GRUB_DEFAULT/cGRUB_DEFAULT=saved' {path_default_grub}"
+    )
+    TestRun.executor.run(f"{grub_version}-set-default {kernel_version}")
+
+
+def _read_kernels(grub_version):
```
Deixx, line 503: rename pls - you are certainly not reading kernels in this method

test_utils/os_utils.py
```diff
@@ -381,3 +380,136 @@ def get_wbt_lat(device: Device):
...
+    versions = TestRun.executor.run(
```
Deixx, line 504: rename - those are grub entries, not versions

test_utils/os_utils.py
```diff
@@ -381,3 +380,136 @@ def get_wbt_lat(device: Device):
...
+        "awk '/menuentry/ && /class/ {count++; print count-1\"#\"$0 }' "
+        f"/boot/{grub_version}/grub.cfg").stdout.splitlines()
+
+    if versions is None:
+        raise CmdException("Error while listing available kernel versions.")
+
+    return versions
+
+
+def read_current_kernel():
```
Deixx, line 514: get_kernel_version_string or sth similar

test_utils/os_utils.py
```diff
@@ -381,3 +380,136 @@ def get_wbt_lat(device: Device):
...
+    )
+    TestRun.executor.run(f"{grub_version}-set-default {kernel_version}")
+
+
+def _read_kernels(grub_version):
+    versions = TestRun.executor.run(
+        "awk '/menuentry/ && /class/ {count++; print count-1\"#\"$0 }' "
+        f"/boot/{grub_version}/grub.cfg").stdout.splitlines()
+
+    if versions is None:
```
Deixx, line 508: Does it ever happen? Even if command fails, you get empty stdout, but not None. splitlines() will still give empty string.

test_utils/os_utils.py
```diff
@@ -381,3 +380,136 @@ def get_wbt_lat(device: Device):
...
+            if main_version not in kernel_version:
+                kernels.append(kernel_version)
+
+    return kernels
+
+
+def _set_kernel_for_next_reboot_only(kernel_version: int, grub_version: str):
+    path_default_grub = "/etc/default/grub"
+    TestRun.executor.run(
+        f"sed -i '/GRUB_DEFAULT/cGRUB_DEFAULT=saved' {path_default_grub}"
```
Deixx, line 490: This will have no effect until you recreate grub.cfg using `grub-mkconf -o <grub.cfg path>`.
Also extract to separate method.

Ostrokrzew: Disagree, I checked it, this command works without using `grub-mkconf -o <grub.cfg path>`.

test_utils/os_utils.py
```diff
@@ -381,3 +380,140 @@ def get_wbt_lat(device: Device):
...
+    available_kernels = _get_entries_from_grub_config(grub_version)
+
+    kernel_version = choose_kernel_version(compatible, available_kernels, custom)
+    if only_1_reboot:
+        _set_kernel_for_next_reboot_only(kernel_version, grub_version)
+    else:
+        _set_default_kernel(kernel_version, grub_version)
+
+
+def choose_kernel_version(compatible: bool, available_kernels: [str], custom: str):
```
Deixx, line 442: add default None value for custom

test_utils/os_utils.py
```diff
@@ -381,3 +380,140 @@ def get_wbt_lat(device: Device):
...
+
+
+def _set_default_kernel(kernel_version: int, grub_version: str):
+    _set_grub_default_entry_to_saved()
+    TestRun.executor.run(f"{grub_version}-set-default {kernel_version}")
+
+
+def _set_grub_default_entry_to_saved():
+    path_default_grub = "/etc/default/grub"
+    TestRun.executor.run(f"sed -i '/GRUB_DEFAULT/cGRUB_DEFAULT=saved' {path_default_grub}")
```
Deixx, line 504: I never said this command will not work, I said it will have no effect until you recreate grub.cfg. /etc/default/grub contains default values used for creating grub.cfg and is used only during that creation. During boot values are read from grub.cfg.

Ostrokrzew: OK, I misunderstood You, I thought your comment refers to single reboot. Of course You're right.

robertbaldyga (approved):

test_tools/fs_utils.py
```diff
@@ -303,3 +314,14 @@ def create_random_test_file(target_file_path: str,
     dd.run()
     file.refresh_item()
     return file
+
+
+def find(search_roots: [str],
+         search_phrase: str,
```
katlapinka, line 320: maybe name it something like "filename_phrase" because my first thought was that is the phrase inside the file, not its name

test_utils/disk_finder.py
```diff
@@ -72,6 +72,8 @@ def discover_ssd_devices(block_devices, devices_res):
     for i in range(0, ssd_count):
         device_path = TestRun.executor.run_expect_success(
             f"isdct show -intelssd {i} | grep DevicePath").stdout.split()[2]
+        if "/dev/sg" in device_path:  # these paths do not have by-id links
+            continue
```
katlapinka, line 76: this change is not needed because I fixed this and other issues with discovering devices in  https://github.com/Open-CAS/test-framework/pull/230

test_utils/os_utils.py
```diff
@@ -226,11 +229,15 @@ def download_file(url, destination_dir="/tmp"):
 
 
 def get_kernel_version():
-    version_string = TestRun.executor.run_expect_success("uname -r").stdout
-    version_string = version_string.split('-')[0]
+    version_string = get_current_kernel_version().split('-')[0]
     return version.Version(version_string)
 
 
+def get_current_kernel_version():
```
katlapinka, line 236: maybe add _str to the name of this method, because those two methods for getting kernel versions have names which could be misleading

katlapinka (changes requested):

Deixx (approved):

karolinavelkaja (approved):

PR #165, opened 2020-06-10
Improvements in TF for perf tests

imjfckm:


test_utils/os_utils.py
```diff
@@ -142,6 +142,25 @@ def get_dut_cpu_number():
...
 
+def get_dut_cpu_physical_cores():
+    """ Get list of CPU numbers that don't share physical cores """
+    output = TestRun.executor.run_expect_success("lscpu --all --parse").stdout
+
+    core_list = []
+    visited_phys_cores = []
+    for line in output.split("\n"):
+        line = line.split("#")[0].rstrip()
+        if len(line) == 0:
```
Ostrokrzew, line 153: Wouldn't
``` python
    if '#' in line:
        continue
```
be easier?

test_tools/fio/fio_param.py
```diff
@@ -303,6 +303,11 @@ def add_job(self, job_name=None):
         self.fio.jobs.append(new_job)
         return new_job
 
+    def clear_jobs(self):
+        self.fio.jobs = []
+
+        return self
```
Ostrokrzew, line 309: Is this `return` necessary?

imjfckm: It's the convention within `FioParam` methods to enable chaining of methods

test_utils/os_utils.py
imjfckm: It would, fixed!

PR #166, opened 2020-06-15
Add rpm module

Ostrokrzew:
Add `rpm` module with basic functionality.

### DIFF
--- a/test_tools/rpm.py
+++ b/test_tools/rpm.py
@@ -1,0 +1,137 @@
+#
+# Copyright(c) 2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+from os.path import join
+from core.test_run import TestRun
+from test_tools.fs_utils import ls, parse_ls_output
+
+
+class Rpm:
+    def __init__(self, name: str, packages_dir: str = None):
+        self.name = name
+        self.packages_dir = packages_dir
+        self.packages_to_install = set()
+        self.update_packages_to_install()
+        self.installed_packages = set()
+        self.update_installed_packages()
+
+    def __str__(self):
+        return self.name
+
+    def make_rpm(self, sources_dir):
+        """
+        Generate installation packages in default directory with 'Make' script.
+        param: sources_dir: directory with make script to generate RPM packages
+        """
+        TestRun.executor.run_expect_success(f"cd {sources_dir} && make rpm")
+        self.packages_dir = join(sources_dir, 'packages')
+
+    def update_packages_to_install(self):
+        """
+        Get list of installation packages from a directory.
+        exception: no packages found
+        """
+        output = parse_ls_output(ls(self.packages_dir), self.packages_dir)
+        if output is not None:
+            self.packages_to_install.update(
+                [file.full_path for file in output if file.full_path.endswith(".rpm")]
+            )
+        else:
+            raise Exception(
+                f"Location {self.packages_dir} doesn't exist. "
+                f"Cannot search there for RPM packages to install."
+            )
+
+    def install_packages(self, force: bool = True):
+        """
+        Install/upgrade packages with the RPM.
+        param: force: ignore warnings related to package conflicts
+        exception: install/upgrade packages failed
+        """
+        TestRun.LOGGER.info("Installing from RPM packages.")
+        if not self.packages_to_install:
+            TestRun.LOGGER.info("No packages to install. Installation skipped.")
+            return
+        cmd = 'rpm --upgrade --hash --verbose '
+        cmd += " ".join(self.packages_to_install)
+        if force:
+            cmd += ' --force'
+
+        try:
+            TestRun.executor.run_expect_success(cmd)
+        except Exception as exc:
+            raise Exception(f"Cannot install/upgrade '{self.packages_to_install}' RPM packages."
+                            f"\n{exc}")
+        self.update_installed_packages()
+
+    def uninstall_packages(self, allmatches: bool = True):
+        """
+        Uninstall packages with the RPM.
+        param: allmatches: remove dependent packages
+        exception: uninstall packages failed
+        """
+        TestRun.LOGGER.info("Uninstalling RPM packages.")
+        cmd = 'rpm --erase --hash --verbose '
+        if self.installed_packages:
+            cmd += " ".join(self.installed_packages)
+        else:
+            cmd += self.name
+        if allmatches:
+            cmd += ' --allmatches'
+
+        try:
+            TestRun.executor.run_expect_success(cmd)
+        except Exception as exc:
+            raise Exception(f"Cannot uninstall '{self.installed_packages}' RPM packages."
+                            f"\n{exc}")
+        self.installed_packages.clear()
+        self.update_packages_to_install()
+
+    def reinstall_packages(self):
+        """
+        Reinstall packages with the RPM.
+        exception: reinstall packages failed
+        """
+        TestRun.LOGGER.info("Reinstalling RPM packages.")
+        if not self.packages_to_install:
+            TestRun.LOGGER.info("No packages to reinstall. Installation skipped.")
+            return
+        cmd = 'rpm --reinstall --hash --verbose '
+        cmd += " ".join(self.packages_to_install)
+
+        try:
+            TestRun.executor.run_expect_success(cmd)
+        except Exception as exc:
+            raise Exception(f"Cannot reinstall {' '.join(self.packages_to_install)} "
+                            f"RPM packages.\n{exc}")
+        self.update_installed_packages()
+
+    def update_installed_packages(self):
+        """
+        Send a query to the RPM to update list of installed packages.
+        """
+        for package in self.packages_to_install:
+            # only package name is correct argument in query, not whole path
+            package = package.split("/")[-1].replace(".rpm", "")
+            cmd = f'rpm --query "{package}"'
+            try:
+                output = TestRun.executor.run_expect_success(cmd)
+                self.installed_packages.update(output.stdout)
+            except Exception:
+                if Rpm.is_package_installed(package):
+                    self.installed_packages.remove(output.stdout)
+                TestRun.LOGGER.info(
+                    f"Cannot retrieve info about installed RPM package '{package}'."
+                )
+
+    @staticmethod
+    def is_package_installed(name):
+        """
+        Check if given package is on the RPM's list of installed packages
+        return: true if package is on the list, false if not
+        """
+        cmd = f'rpm --query "{name}"'
+        output = TestRun.executor.run(cmd)
+        return output.exit_code == 0


test_tools/rpm.py
```diff
@@ -0,0 +1,136 @@
...
+
+from core.test_run import TestRun
+
+
+def make_rpm():
+    """
+    Generate installation packages in default directory.
+    return: path to created packages
+    """
+    TestRun.executor.run(f"cd {TestRun.usr.working_dir} && make rpm")
```
mmichal10, line 14: How about `TestRun.executor.run_expect_success()`?

mmichal10, line 14: I believe you should parameterize this function instead of using ` {TestRun.usr.working_dir}`

test_tools/rpm.py
```diff
@@ -0,0 +1,136 @@
...
+
+def create_rpm_packages(path_to_sources: str, output_dir: str = None):
+    """
+    Use the package generator script directly to generate installation packages.
+    param: path_to_sources: absolute path to sources to build packages
+    param: output_dir: absolute path to output directory where packages will be built
+    return: path to created packages
+    """
+    TestRun.LOGGER.info("Creating OpenCAS RPM packages.")
+    cmd = f'{path_to_sources}/utils/pckgen {path_to_sources} rpm'
```
mmichal10, line 26: Try to avoid fixed paths from OCL repo and think about this module as universal tool for building any rpm

test_tools/rpm.py
```diff
@@ -0,0 +1,136 @@
...
+    param: path_to_packages: absolute path to directory containing wanted packages
+    return: list of absolute paths to packages in given directory
+    exception: no packages found
+    """
+    output = TestRun.executor.run(f"ls -1 {path_to_packages}").stdout.splitlines()
+    if output is not None:
+        packages = []
+        for line in output:
+            packages.append(f'{path_to_packages}/{line}')
+        return packages
```
mmichal10, line 45: `return [f'{path_to_packages}/{line}' for line in output]` looks more pythonic

test_tools/rpm.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    Install packages with the RPM.
+    param: packages: list of installation packages
+    param: force: ignore warnings related to package conflicts
+    return: output from install command
+    """
+    TestRun.LOGGER.info("Installing OpenCAS from RPM packages.")
+    cmd = 'rpm --install'
+    for package in packages:
+        cmd += f' {package}'
```
mmichal10, line 60: How about one line for?

test_tools/rpm.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    Send a query to the RPM to list installed packages filtered by the parameter.
+    param: package_name_filter: installed packages filter
+    return: list of installed packages filtered by the parameter
+    exception: no installed packages found
+    """
+    cmd = 'rpm --query --all'
+    cmd += f' | grep "{package_name_filter}"'
+    packages = TestRun.executor.run(cmd).stdout.splitlines()
+    if any(packages):
```
mmichal10, line 120: You can skip `any()` and use only `if packages:`

test_tools/rpm.py
Ostrokrzew: Can't be done in pretty way.

test_tools/rpm.py
```diff
@@ -0,0 +1,133 @@
...
+
+def get_installed_packages(package_name_filter: str):
+    """
+    Send a query to the RPM to list installed packages filtered by the parameter.
+    param: package_name_filter: installed packages filter
+    return: list of installed packages filtered by the parameter
+    exception: no installed packages found
+    """
+    cmd = 'rpm --query --all'
+    cmd += f' | grep "{package_name_filter}"'
```
robertbaldyga, line 115: What if filter is `None` or an empty string?

test_tools/rpm.py
```diff
@@ -0,0 +1,133 @@
...
+    param: path_to_sources: absolute path to sources to build packages
+    param: output_dir: absolute path to output directory where packages will be built
+    return: path to created packages
+    """
+    TestRun.LOGGER.info("Creating RPM packages.")
+    cmd = f'{path_to_pckgen} {path_to_sources} rpm'
+    cmd += f' -o {output_dir}' if output_dir is not None else ''
+    TestRun.executor.run_expect_success(cmd)
+
+    return output_dir if output_dir is not None else f'{path_to_sources}/packages/'
```
robertbaldyga, line 30: The `pckgen` is OCL specific utility script. I don't feel that is should be wrapped in TF.

robertbaldyga (dismissed):

test_tools/rpm.py
Ostrokrzew: good point

test_tools/rpm.py
```diff
@@ -0,0 +1,118 @@
...
+def get_packages_from_dir(path_to_packages: str):
+    """
+    Get list of installation packages from a directory.
+    param: path_to_packages: absolute path to directory containing wanted packages
+    return: list of absolute paths to packages in given directory
+    exception: no packages found
+    """
+    output = TestRun.executor.run(f"ls -1 {path_to_packages}").stdout.splitlines()
+    if output is not None:
+        return [f'{path_to_packages}/{line}' for line in output]
```
mmichal10, line 27: How about using `os.path.join()`?

test_tools/rpm.py
```diff
@@ -0,0 +1,118 @@
...
+from core.test_run import TestRun
+
+
+def make_rpm(path_to_make: str):
+    """
+    Generate installation packages in default directory with 'Make' script.
+    return: path to created packages
+    """
+    TestRun.executor.run_expect_success(f"cd {path_to_make} && make rpm")
+    return f'{path_to_make}/packages/'
```
mmichal10, line 15: How about using `os.path.join()`?

test_tools/rpm.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+    Uninstall packages with the RPM.
+    param: packages: list of installed packages to remove
+    param: allmatches: remove dependent packages
+    return: output from uninstall command
+    """
+    TestRun.LOGGER.info("Uninstalling RPM packages.")
+    cmd = 'rpm --erase'
+    for package in packages:
+        cmd += f' {package}'
```
rafalste, line 59: Instead of filenames, you need to supply the 'rpm --erase' command with packages names as listed in 'rpm -qa' output. Function 'get_installed_packages()' from below will do the work. Its output will be perfect here. In case you want to uninstall only one version of package and not all that matches (if you installed new version on top of the old one with --install --force before), you need to collect packages version with, for example `rpm -qp --queryformat "%{VERSION}" <rpm_filename>` during the installation step.

test_tools/rpm.py
```diff
@@ -0,0 +1,119 @@
...
+
+def install_packages(packages: list, force: bool = True):
+    """
+    Install packages with the RPM.
+    param: packages: list of installation packages
+    param: force: ignore warnings related to package conflicts
+    return: output from install command
+    """
+    TestRun.LOGGER.info("Installing from RPM packages.")
+    cmd = 'rpm --install'
```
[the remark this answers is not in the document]
Ostrokrzew: I think it is redundant to add `--hash` and `--verbose` options, but maybe I'm wrong.
You can find upgrade below and it takes package files as arguments (like RPM's help says).
In this case I want to install packages, no matter if they're already installed (they're not, I'm uninstalling them and tests would do the same).

test_tools/rpm.py
Ostrokrzew: I think it's not needed for tests to remove the particular version of a package, because @mmichal10 introduced the installation of a custom CAS's version at the setup phase. In my opinion it is better to have a clean installation to test, with no conflicts.

test_tools/rpm.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def get_packages_from_dir(path_to_packages: str):
+    """
+    Get list of installation packages from a directory.
+    param: path_to_packages: absolute path to directory containing wanted packages
+    return: list of absolute paths to packages in given directory
+    exception: no packages found
+    """
+    output = TestRun.executor.run(f"ls -1 {path_to_packages}").stdout.splitlines()
```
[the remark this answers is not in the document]
Ostrokrzew: I don't know (yet) why, but `os.listdir(path_to_packages)` raises 'FileNotFoundError: [Errno 2] No such file or directory'. I'll investigate this.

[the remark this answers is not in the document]
katlapinka: Rafal's solution will work only on test controller for remote test execution, not on the DUT. But Slawek, you can use methods from fs_utils - there is a method for listing dir which returns objects list

test_tools/rpm.py
```diff
@@ -0,0 +1,123 @@
...
+        self.packages_dir = join(cas_sources_dir, 'packages')
+
+    def get_packages_from_dir(self):
+        """
+        Get list of installation packages from a directory.
+        param: path_to_packages: absolute path to directory containing wanted packages
+        return: list of absolute paths to packages in given directory
+        exception: no packages found
+        """
+        output = TestRun.executor.run(f"ls -1 {self.packages_dir}").stdout.splitlines()
```
katlapinka, line 34: you can use method from fs_utils - there is ls method for listing dir and parse_ls_output method which returns specific types objects list

test_tools/rpm.py
```diff
@@ -0,0 +1,124 @@
...
+        self.packages_dir = packages_dir
+        self.packages_to_install = set()
+        self.get_packages_from_dir()
+        self.installed_packages = set()
+        self.get_installed_packages()
+
+    def make_rpm(self, cas_sources_dir):
+        """
+        Generate installation packages in default directory with 'Make' script.
+        return: path to created packages
```
mmichal10, line 23: Function doesn't return any value

test_tools/rpm.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        TestRun.executor.run_expect_success(f"cd {cas_sources_dir} && make rpm")
+        self.packages_dir = join(cas_sources_dir, 'packages')
+
+    def get_packages_from_dir(self):
+        """
+        Get list of installation packages from a directory.
+        param: path_to_packages: absolute path to directory containing wanted packages
+        return: list of absolute paths to packages in given directory
```
mmichal10, line 32: Function doesn't return any value, but if it's name starts `get_` it could. Otherwise you should consider renaming it to sth like `update_`

test_tools/rpm.py
```diff
@@ -0,0 +1,124 @@
...
+                f"Location {self.packages_dir} doesn't exist. "
+                f"Cannot search there for RPM packages to install."
+            )
+
+    def install_packages(self, force: bool = True):
+        """
+        Install/upgrade packages with the RPM.
+        param: packages: list of installation packages
+        param: force: ignore warnings related to package conflicts
+        return: output from install command
```
mmichal10, line 51: Function doesn't return any value

test_tools/rpm.py
```diff
@@ -0,0 +1,124 @@
...
+            self.get_installed_packages()
+        else:
+            raise Exception(f"Cannot install/upgrade '{self.packages_to_install}' RPM packages.")
+
+    def uninstall_packages(self, allmatches: bool = True):
+        """
+        Uninstall packages with the RPM.
+        param: packages: list of installed packages to remove
+        param: allmatches: remove dependent packages
+        return: output from uninstall command
```
mmichal10, line 70: Function doesn't return any value

test_tools/rpm.py
```diff
@@ -0,0 +1,124 @@
...
+        if TestRun.executor.run(cmd).exit_code == 0:
+            self.installed_packages.clear()
+        else:
+            raise Exception(f"Cannot uninstall '{self.installed_packages}' RPM packages.")
+
+    def reinstall_packages(self):
+        """
+        Reinstall packages with the RPM.
+        param: packages: list of installed packages to reinstall
+        return: output from reinstall command
```
mmichal10, line 87: Function doesn't return any value

test_tools/rpm.py
```diff
@@ -0,0 +1,124 @@
...
+        if TestRun.executor.run(cmd).exit_code == 0:
+            self.get_installed_packages()
+        else:
+            raise Exception(f"Cannot reinstall '{self.installed_packages}' RPM packages.")
+
+    def get_installed_packages(self):
+        """
+        Send a query to the RPM to list installed packages filtered by the parameter.
+        param: package_name_filter: installed packages filter
+        return: list of installed packages filtered by the parameter
```
mmichal10, line 103: Function doesn't return any value, but if it's name starts `get_` it could. Otherwise you should consider renaming it to sth like `update_`

test_tools/rpm.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def update_packages_to_install(self):
+        """
+        Get list of installation packages from a directory.
+        exception: no packages found
+        """
+        output = parse_ls_output(ls(self.packages_dir), self.packages_dir)
+        if output is not None:
+            self.packages_to_install.update(
+                [file.full_path if file.full_path.endswith(".rpm") else None] for file in output
```
Deixx, line 36: If you're doing list comprehension, then the ']' should be at the end of the line.
Additionally we probably don't need None added to the set for non-rpm files, so I'd suggest:
```python
[file.full_path for file in output if file.full_path.endswith(".rpm")]
```

test_tools/rpm.py
```diff
@@ -0,0 +1,115 @@
...
+        self.packages_dir = packages_dir
+        self.packages_to_install = set()
+        self.update_packages_to_install()
+        self.installed_packages = set()
+        self.update_installed_packages()
+
+    def make_rpm(self, cas_sources_dir):
+        """
+        Generate installation packages in default directory with 'Make' script.
+        param: cas_sources_dir: directory with make script to generate RPM packages
```
Deixx, line 23: This tool doesn't seem to be restricted only to CAS packages, so you should probably lose 'cas' from param name

test_tools/rpm.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        Install/upgrade packages with the RPM.
+        param: force: ignore warnings related to package conflicts
+        exception: install/upgrade packages failed
+        """
+        TestRun.LOGGER.info("Installing from RPM packages.")
+        cmd = 'rpm --upgrade --hash --verbose'
+        for package in self.packages_to_install:
+            cmd += f' {package}'
+        cmd += ' --force' if force else ''
```
Deixx, line 54: I know it's a nice one-liner, but you're actually not adding anything in else, so I'd suggest:
```python
if force:
    cmd += ' --force'
```

test_tools/rpm.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        Uninstall packages with the RPM.
+        param: allmatches: remove dependent packages
+        exception: uninstall packages failed
+        """
+        TestRun.LOGGER.info("Uninstalling RPM packages.")
+        cmd = 'rpm --erase --hash --verbose'
+        for package in self.installed_packages:
+            cmd += f' {package}'
+        cmd += ' --allmatches' if allmatches else ''
```
Deixx, line 72: another "empty" else

test_tools/rpm.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def uninstall_packages(self, allmatches: bool = True):
+        """
+        Uninstall packages with the RPM.
+        param: allmatches: remove dependent packages
+        exception: uninstall packages failed
+        """
+        TestRun.LOGGER.info("Uninstalling RPM packages.")
+        cmd = 'rpm --erase --hash --verbose'
+        for package in self.installed_packages:
```
Deixx, line 70:
```python
cmd += " " + " ".join(self.installed_packages)
```
or the same comment as before ;)

test_tools/rpm.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def install_packages(self, force: bool = True):
+        """
+        Install/upgrade packages with the RPM.
+        param: force: ignore warnings related to package conflicts
+        exception: install/upgrade packages failed
+        """
+        TestRun.LOGGER.info("Installing from RPM packages.")
+        cmd = 'rpm --upgrade --hash --verbose'
+        for package in self.packages_to_install:
```
Deixx, line 52:
```python
cmd += " " + " ".join(self.packages_to_install)
```
or you can add space after '--verbose' in initial cmd

test_tools/rpm.py
```diff
@@ -0,0 +1,115 @@
...
+            raise Exception(f"Cannot uninstall '{self.installed_packages}' RPM packages.")
+
+    def reinstall_packages(self):
+        """
+        Reinstall packages with the RPM.
+        exception: reinstall packages failed
+        """
+        TestRun.LOGGER.info("Reinstalling RPM packages.")
+        cmd = 'rpm --reinstall --hash --verbose'
+        for package in self.installed_packages:
```
Deixx, line 86:
```python
cmd += " ".join(self.installed_packages)
```
or...

test_tools/rpm.py
```diff
@@ -0,0 +1,115 @@
...
+            self.update_installed_packages()
+        else:
+            raise Exception(f"Cannot reinstall '{self.installed_packages}' RPM packages.")
+
+    def update_installed_packages(self):
+        """
+        Send a query to the RPM to list installed packages filtered by the parameter.
+        """
+        cmd = 'rpm --query --all'
+        cmd += f' | grep "{self.name}"'
```
Deixx, line 99: Why split cmd to 2 lines?
Also grep "name" will list everything containing name, regardless if it's related to actual package or not

test_tools/rpm.py
```diff
@@ -0,0 +1,115 @@
...
+        output = TestRun.executor.run(cmd)
+        if output.exit_code == 0:
+            self.installed_packages.update(output.stdout.splitlines())
+        else:
+            TestRun.LOGGER.info(f"Cannot retrieve info about installed '{self.name}' RPM packages.")
+
+    def is_package_installed(self):
```
Deixx, line 106: I'd make it a static method that queries system for installed RPMs and checks if list contains name (passed as argument)

test_tools/rpm.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        Check if given package is on the RPM's list of installed packages
+        return: true if package is on the list, false if not
+        """
+        for package in self.installed_packages:
```
Deixx, line 111:
```python
return any(map(lambda package: self.name in package, self.installed_packages))
```
Of course name could be in package which is not necessary the package we search for

test_tools/rpm.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        Check if given package is on the RPM's list of installed packages
+        return: true if package is on the list, false if not
+        """
+        cmd = f'rpm --query "{name}"'
+        output = TestRun.executor.run(cmd)
+        if output.exit_code == 0:
+            return True
+        else:
+            return False
```
Deixx, line 129: return output.exit_code == 0

Ostrokrzew: right

Deixx (dismissed):

mmichal10 (dismissed):

test_tools/rpm.py
```diff
@@ -0,0 +1,126 @@
...
+        TestRun.LOGGER.info("Installing from RPM packages.")
+        cmd = 'rpm --upgrade --hash --verbose '
+        cmd += " ".join(self.packages_to_install)
+        if force:
+            cmd += ' --force'
+
+        if TestRun.executor.run(cmd).exit_code == 0:
+            self.update_installed_packages()
+        else:
+            raise Exception(f"Cannot install/upgrade '{self.packages_to_install}' RPM packages.")
```
rafalste, line 62: Maybe test first if packages_to_install is not empty before running this code? If it will be empty there will be a message that `Cannot install/upgrade '' RPM packages`, but that may be confusing.
Tell me if I'm just nitpicking. ;)

test_tools/rpm.py
```diff
@@ -0,0 +1,126 @@
...
+            self.update_installed_packages()
+        else:
+            raise Exception(f"Cannot reinstall {' '.join(self.packages_to_install)} "
+                            f"RPM packages.")
+
+    def update_installed_packages(self):
+        """
+        Send a query to the RPM to update list of installed packages.
+        """
+        for package in self.packages_to_install:
```
rafalste, line 104: What if packages_to_install is empty (in case there were no RPMs in the given location or you don't want to install anything so you don't pass any packages)? Maybe it will be a better idea to search for installed packages based on self.name instead? That way you can query for packages only once to return all of them and get rid of the for loop.
You can do this by slightly modifying your check to list all installed packages and filter them using grep:
`rpm -qa | grep <name>`

test_tools/rpm.py
```diff
@@ -0,0 +1,126 @@
...
+        """
+        TestRun.executor.run_expect_success(f"cd {sources_dir} && make rpm")
+        self.packages_dir = join(sources_dir, 'packages')
+
+    def update_packages_to_install(self):
+        """
+        Get list of installation packages from a directory.
+        exception: no packages found
+        """
+        output = parse_ls_output(ls(self.packages_dir), self.packages_dir)
```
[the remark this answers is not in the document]
Ostrokrzew: The `ls()` function would return empty string and the `parse_ls_output()` would return empty list, but not None object. I checked it because I'm creating the `rpm` object in the `installer` module with only one argument and then calling its `make_rpm()` method to initialize `packages_to_install`. It wouldn't search for any packages in working directory.

test_tools/rpm.py
Ostrokrzew: It will just skip querying. Similar issue was discussed above with Deixx (https://github.com/Open-CAS/test-framework/pull/166#discussion_r469395166) and I resigned from version with `grep`.

test_tools/rpm.py
```diff
@@ -0,0 +1,132 @@
...
+        TestRun.LOGGER.info("Installing from RPM packages.")
+        if not self.packages_to_install:
+            TestRun.LOGGER.info("No packages to install. Installation skipped.")
+            return
+        cmd = 'rpm --upgrade --hash --verbose '
+        cmd += " ".join(self.packages_to_install)
+        if force:
+            cmd += ' --force'
+
+        if TestRun.executor.run(cmd).exit_code == 0:
```
robertbaldyga, line 62: `TestRun.executor.run_expect_success()` would be more idiomatic.

test_tools/rpm.py
```diff
@@ -0,0 +1,132 @@
...
+        TestRun.LOGGER.info("Uninstalling RPM packages.")
+        cmd = 'rpm --erase --hash --verbose '
+        if self.installed_packages:
+            cmd += " ".join(self.installed_packages)
+        else:
+            cmd += self.name
+        if allmatches:
+            cmd += ' --allmatches'
+
+        if TestRun.executor.run(cmd).exit_code == 0:
```
robertbaldyga, line 82: Ditto.

test_tools/rpm.py
```diff
@@ -0,0 +1,132 @@
...
+        exception: reinstall packages failed
+        """
+        TestRun.LOGGER.info("Reinstalling RPM packages.")
+        if not self.packages_to_install:
+            TestRun.LOGGER.info("No packages to reinstall. Installation skipped.")
+            return
+        cmd = 'rpm --reinstall --hash --verbose '
+        cmd += " ".join(self.packages_to_install)
+
+        if TestRun.executor.run(cmd).exit_code == 0:
```
robertbaldyga, line 100: Ditto.

test_tools/rpm.py
```diff
@@ -0,0 +1,132 @@
...
+    def update_installed_packages(self):
+        """
+        Send a query to the RPM to update list of installed packages.
+        """
+        for package in self.packages_to_install:
+            # only package name is correct argument in query, not whole path
+            package = package.split("/")[-1].replace(".rpm", "")
+            cmd = f'rpm --query "{package}"'
+            output = TestRun.executor.run(cmd)
+            if output.exit_code == 0:
```
robertbaldyga, line 115: Ditto.

test_tools/rpm.py
```diff
@@ -0,0 +1,132 @@
...
+
+    @staticmethod
+    def is_package_installed(name):
+        """
+        Check if given package is on the RPM's list of installed packages
+        return: true if package is on the list, false if not
+        """
+        cmd = f'rpm --query "{name}"'
+        output = TestRun.executor.run(cmd)
+        return output.exit_code == 0
```
robertbaldyga, line 132: Ditto.

Ostrokrzew: Agree to above, but it is simple true/ false test and IMO it doesn't make sense to implement try-catch mechanism here. I don't want to raise exception when I'm asking machine if there's some package installed and the answer is 'no'. Please, tell me if I'm wrong.

robertbaldyga: Ok, that sounds reasonable.

robertbaldyga (approved):

Deixx (approved):

PR #171, opened 2020-06-23
add begin & end add to partition and other some minor fixes

micrakow:


test_utils/size.py
```diff
@@ -73,6 +73,22 @@ def __init__(self, unit):
...
+            self.MebiByte: 'MiB',
+            self.GibiByte: 'GiB',
+            self.TebiByte: 'TiB',
+            self.KiloByte: 'kB',
+            self.MegaByte: 'MB',
+            self.GigaByte: 'GB',
+            self.TeraByte: 'TB'
+        }
+        return unit_string.get(self, "Invalid unit.")
+
```
katlapinka, line 91: shouldn't it be in Size class?

test_tools/disk_utils.py
```diff
@@ -80,20 +80,21 @@ def create_partition(
...
         f"Creating {part_type.name} partition on device: {device.system_path}")
 
-    begin = get_first_partition_offset(device, aligned).get_value(unit)
+    begin = get_first_partition_offset(device, aligned)
     for part in device.partitions:
-        begin += part.size.get_value(unit)
+        begin += part.size
         if part.type == PartitionType.logical:
-            begin += Size(1, Unit.MebiByte if not aligned else device.block_size).get_value(unit)
+            begin += Size(1, Unit.MebiByte if not aligned else device.block_size)
```
mmichal10, line 87: Since `Partition` class has `end` attribute, wouldn't it be more readable to find address You are looking for using this loop?

micrakow: Let's leave it as it is as we discussed for now

PR #175, opened 2020-07-08
Add runlevels API

katlapinka:

### DIFF
--- a/test_utils/os_utils.py
+++ b/test_utils/os_utils.py
@@ -7,7 +7,7 @@ import time
 from datetime import timedelta, datetime
 import os
 
-from aenum import IntFlag, Enum
+from aenum import IntFlag, Enum, IntEnum
 from packaging import version
 
 from storage_devices.device import Device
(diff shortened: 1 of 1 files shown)


test_utils/os_utils.py
```diff
@@ -25,6 +25,103 @@ class DropCachesMode(IntFlag):
...
+        else:
+            TestRun.executor.run_expect_success(
+                f"sed -i 's/^.*id:.*$/id:{runlevel.value}:initdefault: /' /etc/inittab")
+    TestRun.executor.run_expect_success(f"init {runlevel.value}")
+
+
+def get_runlevel():
+    result = TestRun.executor.run("runlevel")
+    split_output = result.stdout.split()
+    if len(split_output) == 2:
```
robertbaldyga, line 115: We may want to check exit code of the command. This will be better indicator of success that number of columns in stdout. `run_expect_success()` should do the job.

test_utils/os_utils.py
```diff
@@ -25,6 +25,103 @@ class DropCachesMode(IntFlag):
...
+        try:
+            runlevel = Runlevel(int(split_output[1]))
+            return runlevel
+        except Exception:
+            raise Exception(f"Cannot parse '{split_output[1]}' to runlevel.")
+    else:
+        return Runlevel.unknown
```
robertbaldyga, line 122: Is it useful? Don't we want to have this `unknown` value instead of just raising exception?

test_utils/os_utils.py
```diff
@@ -25,6 +25,103 @@ class DropCachesMode(IntFlag):
...
+        return SystemManagerType.sysv
+    elif type == "systemd":
+        return SystemManagerType.systemd
+    raise Exception(f"Unknown system manager type ({type}).")
+
+
+def change_runlevel(runlevel: Runlevel):
+    if runlevel == get_runlevel():
+        return
+    if 0 < runlevel.value < 6:
```
robertbaldyga, line 102: Why not compare with enum values?

PR #180, opened 2020-07-20
Add pattern utils for FIO

Ostrokrzew:


test_tools/fio/fio_patterns.py
```diff
@@ -0,0 +1,45 @@
...
+    for i in range(0, 256):
+        seq_pattern += f"{hex(i).lstrip('0x'):0>2}"
+    return seq_pattern.strip()
+
+
+def _get_random_pattern():
+    random_pattern = "0x"
+    for i in random.sample(range(256), random.randint(1, 256)):
+        random_pattern += f"{hex(i).lstrip('0x'):0>2}"
+    return random_pattern.strip()
```
katlapinka, line 21: You can move those methods to Pattern class

test_tools/fio/fio_patterns.py
```diff
@@ -0,0 +1,42 @@
...
+    bin_2 = "0x55"
+    random = _set_random_pattern()
+
+    def get_unique_patterns():
+        return set(list(Pattern.__members__.values()))
+
+    def mix_patterns():
+        patterns = list(Pattern.get_unique_patterns())
+        random.shuffle(patterns)
+        return patterns
```
robertbaldyga, line 39: This doesn't look like generally useful functionality. I think it should be moved to the test which uses it.

test_tools/fio/fio_patterns.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def __str__(self):
+        return self.value
```
robertbaldyga, line 42: Do we really need to override default method? This is quite a standard to obtain enum value by `Class.item.value` property.

test_tools/fio/fio_patterns.py
```diff
@@ -0,0 +1,42 @@
...
+import random
+from aenum import Enum
+
+
+class Pattern(Enum):
+    def _set_seq_pattern():
+        seq_pattern = "0x"
+        for i in range(0, 256):
+            seq_pattern += f"{hex(i).lstrip('0x'):0>2}"
+        return seq_pattern.strip()
```
robertbaldyga, line 15:
```python
return "0x"+"".join([f"{i:02x}" for i in range(0, 256)])
```

test_tools/fio/fio_patterns.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def _set_random_pattern():
+        random_pattern = "0x"
+        for i in random.sample(range(256), random.randint(1, 256)):
+            random_pattern += f"{hex(i).lstrip('0x'):0>2}"
+        return random_pattern.strip()
```
robertbaldyga, line 21:
```python
return "0x"+secrets.token_hex(random.randint(1, 256))
```
BTW do we really want this pattern to be variable length?

Ostrokrzew: nope, removed

test_tools/fio/fio_patterns.py
Ostrokrzew: thx

robertbaldyga (approved):

PR #186, opened 2020-07-31
Simple API for VDBench

katlapinka:

### DIFF
--- a/internal_plugins/vdbench/__init__.py
+++ b/internal_plugins/vdbench/__init__.py
@@ -1,0 +1,96 @@
+#
+# Copyright(c) 2020 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+
+import os
+import time
+from datetime import timedelta
+from core.test_run import TestRun
+from test_tools import fs_utils
+
+
+class Vdbench:
+    def __init__(self, params, config):
+        print("VDBench plugin initialization")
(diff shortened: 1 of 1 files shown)


test_tools/vdbench.py
```diff
@@ -0,0 +1,74 @@
...
+import time
+from datetime import timedelta
+from core.test_run import TestRun
+from test_tools import fs_utils
+
+
+class Vdbench:
+    def __init__(self, reinstall=False):
+        self.run_time = timedelta(seconds=60)
+        self.working_dir = TestRun.config["vdbench_working_dir"]
```
robertbaldyga, line 16: This looks like a good candidate for internal plugin. That way it will have it's own namespace in config, so that we will not introduce new global property, which BTW should be optional (as not every user will have vdbench in his environment), so making it a plugin makes even more sense.

internal_plugins/vdbench/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+import time
+from datetime import timedelta
+from core.test_run import TestRun
+from test_tools import fs_utils
+
+
+class Vdbench:
+    def __init__(self, params, config):
+        print("VDBench plugin initialization")
+        self.config = config
```
robertbaldyga, line 16: Mandatory config properties should be validated in constructor.

Ostrokrzew (approved):

PR #191, opened 2020-08-06
Implement sparse parametrization (pair-testing)

imjfckm:
Introduce new parametrization mark (parametrizex) which will respond
to --parametrization-type pytest argument. This mark supports providing
single parameter name and single parameter values iterable for now.
Parametrize's multiple params per mark and marking/skipping individual
testcases is not supported

If "full" is used then standard parametrization scheme is engaged
(cartesian product of all argument values).

"pair" parametrization scheme is ensuring all parameter value pairs are
present in tested cases. This includes some randomization to ensure
greater coverage with multiple runs.

### DIFF
--- a/core/test_run_utils.py
+++ b/core/test_run_utils.py
@@ -4,8 +4,9 @@
 #
 
 
-import pytest
 from IPy import IP
+import traceback
+import pytest
 
 from connection.ssh_executor import SshExecutor
 from connection.local_executor import LocalExecutor
@@ -13,9 +14,9 @@ from storage_devices.disk import Disk
 from test_utils import disk_finder
 from test_utils.dut import Dut
 from core.plugins import PluginManager
+from core.pair_testing import generate_pair_testing_testcases, register_testcases
 from log.base_log import BaseLogResult
 import core.test_run
-import traceback
 
 
 TestRun = core.test_run.TestRun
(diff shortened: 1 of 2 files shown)


core/test_run_utils.py
```diff
@@ -160,6 +165,41 @@ def __makereport(cls, item, call, res):
...
+    else:
+        raise Exception("Not supported parametrization type")
+
+
+TestRun.generate_tests = __generate_tests
+
+
+@classmethod
+def __addoption(cls, parser):
+    parser.addoption("--parametrization-type", choices=["pair", "full"], default="full")
```
robertbaldyga, line 197: IMO we should make "pair" the default to avoid time consuming mistakes.

core/test_run_utils.py
```diff
@@ -160,6 +165,41 @@ def __makereport(cls, item, call, res):
...
+    argnames = []
+    argvals = []
+    for mark in parametrizex_marks:
+        argnames.append(mark.args[0])
+        argvals.append(list(mark.args[1]))
+
+    if metafunc.config.getoption("--parametrization-type") == "full":
+        for name, values in zip(argnames, argvals):
+            metafunc.parametrize(name, values)
+        return
```
robertbaldyga, line 183: This `return` is noop here.

core/test_run_utils.py
imjfckm: Done

core/test_run_utils.py
imjfckm: Whoops, fixed

PR #192, opened 2020-08-10
RAID API

Deixx:


test_tools/mdadm.py
```diff
@@ -0,0 +1,87 @@
...
+        return " ".join([d.system_path for d in devices])
+    except TypeError:  # single device provided instead of a list
+        return devices.system_path
+    except AttributeError:  # string provided instead of devices
+        return devices
+
+
+class Mdadm:
+    @staticmethod
+    def assemble(devices: [Device] = None):
```
robertbaldyga, line 24: IMO low level tool wrappers shouldn't use higher level abstraction classes like `Device`. It creates interdependencies between components, while from design perspective one-way dependencies is preferred. Passing simple list of device paths should be enough in this case. The same applies to other methods.

storage_devices/raid.py
```diff
@@ -0,0 +1,230 @@
...
+
+    @staticmethod
+    def discover():
+        TestRun.LOGGER.info("Discover RAIDs in system...")
+        raids = []
+
+        # sometimes links for RAIDs are not properly created, force udev to create them
+        TestRun.executor.run("udevadm trigger")
+
+        mdadm_scan = Mdadm.examine().stdout
```
robertbaldyga, line 108: Why this output is not parsed in `Mdadm`? It's responsible for interfacing with mdadm utility after all, so parsing output to more useful form seems to be right task for that class.

storage_devices/raid.py
```diff
@@ -0,0 +1,230 @@
...
+
+from core.test_run import TestRun
+from storage_devices.device import Device
+from storage_devices.disk import Disk
+from test_tools.mdadm import Mdadm
+from test_utils.output import CmdException
+from test_utils.size import Size
+
+readlink_uuid_cmd = "readlink --canonicalize-existing /dev/disk/by-id/md-uuid-"
+readlink_md_cmd = "readlink --canonicalize-existing /dev/md/"
```
robertbaldyga, line 16: We use readlink also in few other places, so maybe it already deserves its own utility wrapper.

storage_devices/raid.py
```diff
@@ -0,0 +1,230 @@
...
+
+    @staticmethod
+    def _get_level(details: str):
+        level = [line for line in details.splitlines() if "Raid Level" in line][0].split(" : ")[-1]
+        return Level[level.capitalize()]
+
+    @staticmethod
+    def _get_uuid(details: str):
+        uuid = [line for line in details.splitlines() if "UUID" in line][0].split(" : ")[-1]
+        return uuid
```
robertbaldyga, line 224: This parsing also could be made inside `Mdadm` class.

test_tools/fs_utils.py
```diff
@@ -281,6 +281,12 @@ def parse_ls_output(ls_output, dir_path=''):
     return fs_items
 
 
+def readlink(link: str, options="--canonicalize-existing"):
+    return TestRun.executor.run_expect_success(
+        f"readlink {options} {link}"
+    ).stdout
```
Ostrokrzew, line 287: Maybe move it to Symlink class (test_utils/filesystem/symlink.py)?

Deixx: Symlink is a class designed to represent actual object, readlink is just a system utility

test_tools/mdadm.py
```diff
@@ -0,0 +1,146 @@
...
+        if conf.metadata.value != "legacy":
+            cmd += f"--metadata={conf.metadata.value} "
+        if conf.level is not None:
+            cmd += f"--level={conf.level.value} "
+        if conf.number_of_devices:
+            cmd += f"--raid-devices={conf.number_of_devices} "
+        if conf.strip_size:
+            cmd += f"--chunk={conf.strip_size} "
+        if conf.size:
+            cmd += f"--size={conf.size.get_value(Unit.KibiByte)} "
```
Ostrokrzew, line 36: Please, cast this to int.

Ostrokrzew (changes requested):

storage_devices/raid.py
```diff
@@ -0,0 +1,181 @@
...
+            if container_output.exit_code != 0:
+                raise CmdException(f"Failed to start {raid_conf.metadata.value} container",
+                                   container_output)
+
+        if not raid_conf.name:
+            raid_conf.name = cls.__get_unique_name()
+
+        raid_output = Mdadm.create(raid_conf, get_devices_paths_string(volume_devices))
+        if raid_output.exit_code != 0:
+            raise CmdException(f"Failed to start RAID volume", raid_output)
```
robertbaldyga, line 157: We could raise this exception inside `Mdadm.create()` instead of returning exit code.

storage_devices/raid.py
```diff
@@ -0,0 +1,181 @@
...
+    def __eq__(self, other):
+        try:
+            return self.uuid == other.uuid
+        except AttributeError:
+            return False
+
+    def stop(self):
+        Mdadm.stop(self.system_path)
+        if self.container_path:
+            Mdadm.stop(self.container_path)
```
robertbaldyga, line 99: Missing check for potential errors (btw `Mdadm.stop()` could also raise an exception).

Ostrokrzew (approved):

robertbaldyga (approved):

PR #194, opened 2020-08-11
Reinitialize random number generator for reproducibility

imjfckm:

### DIFF
--- a/core/test_run_utils.py
+++ b/core/test_run_utils.py
@@ -7,6 +7,8 @@
 from IPy import IP
 import traceback
 import pytest
+import random
+import sys
 
 from connection.ssh_executor import SshExecutor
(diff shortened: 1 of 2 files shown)


core/test_run_utils.py
```diff
@@ -41,9 +42,12 @@ def __configure(cls, config):
         "os_dependent: run test only if its OS dependent, otherwise skip"
     )
     config.addinivalue_line(
+        "markers",
         "parametrizex(argname, argvalues): sparse parametrized testing"
     )
 
+    cls.random_seed = config.getoption("--random-seed") or int(random.random() * 2**32)
```
robertbaldyga, line 49: `random.randrange(sys.maxsize)` would look better to me.

imjfckm: Fixed

PR #200, opened 2020-08-17
Power control libvirt plugin config check

katlapinka:

### DIFF
--- a/internal_plugins/power_control_libvirt/__init__.py
+++ b/internal_plugins/power_control_libvirt/__init__.py
@@ -2,7 +2,7 @@
 # Copyright(c) 2020 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 #
-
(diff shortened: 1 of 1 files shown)


internal_plugins/power_control_libvirt/__init__.py
```diff
@@ -11,6 +11,8 @@ class PowerControlPlugin:
     def __init__(self, params, config):
         print("Power Control LibVirt Plugin initialization")
         self.config = config
+        if not all(name in self.config for name in ["ip", "user", "password"]):
+            raise Exception("Missing fields in config! (ip, user and password required)")
```
robertbaldyga, line 15: Maybe it would be better to assign config properties to class fields? That way we can be sure they are all present and that nobody is going to silently introduce new config property without validating it in constructor.

Ostrokrzew (approved):

PR #201, opened 2020-08-17
Set scsi_debug config as a field in class

katlapinka:


internal_plugins/scsi_debug/__init__.py
```diff
@@ -14,6 +14,7 @@ class ScsiDebug:
     def __init__(self, params, config):
         self.params = params
         self.module_name = "scsi_debug"
+        self.config = config
```
robertbaldyga, line 17: Why should we do this if none method of this class uses this config?

PR #212, opened 2020-09-17
Make timeout for power cycle same as for reboot

Ostrokrzew:

### DIFF
--- a/internal_plugins/power_control_libvirt/__init__.py
+++ b/internal_plugins/power_control_libvirt/__init__.py
@@ -2,6 +2,8 @@
 # Copyright(c) 2020 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 #
+from datetime import timedelta
+
 from connection.local_executor import LocalExecutor
 from connection.ssh_executor import SshExecutor
 from core.test_run import TestRun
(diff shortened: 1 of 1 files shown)


connection/base_executor.py
```diff
@@ -12,6 +12,9 @@
 
 
 class BaseExecutor:
+    def __init__(self):
+        self.reboot_timeout = None
+
```
robertbaldyga, line 17: Reboot is ability of `SshExecutor`. It should not be in the `BaseExecutor`.

internal_plugins/power_control_libvirt/__init__.py
```diff
@@ -38,7 +40,8 @@ def teardown(self):
     def power_cycle(self):
         self.executor.run(f"virsh reset {self.config['domain']}")
         TestRun.executor.wait_for_connection_loss()
-        TestRun.executor.wait_for_connection()
+        (TestRun.executor.wait_for_connection(timedelta(seconds=TestRun.executor.reboot_timeout))
+         if TestRun.executor.reboot_timeout is not None else TestRun.executor.wait_for_connection())
```
robertbaldyga, line 44: You can obtain reboot timeout value from `TestRun.config`. That way you don't need to access executor internals.

robertbaldyga (changes requested):

internal_plugins/power_control_libvirt/__init__.py
```diff
@@ -38,7 +40,11 @@ def teardown(self):
     def power_cycle(self):
         self.executor.run(f"virsh reset {self.config['domain']}")
         TestRun.executor.wait_for_connection_loss()
-        TestRun.executor.wait_for_connection()
+        timeout = TestRun.config.get('reboot_timeout')
+        if timeout:
+            TestRun.executor.wait_for_connection(timedelta(seconds=timeout))
```
robertbaldyga, line 45:
```python
timedelta(seconds=int(timeout))
```
`timedelta` doesn't accept `str`

robertbaldyga (changes requested):

Deixx (approved):

katlapinka (approved):

robertbaldyga (approved):

PR #218, opened 2020-11-25
fix for remove_partitions method

karolinavelkaja:
Signed-off-by: Karolina Rogowska <karolina.rogowsky@gmail.com>

### DIFF
--- a/test_tools/disk_utils.py
+++ b/test_tools/disk_utils.py
@@ -232,6 +232,7 @@ def remove_partitions(device):
         .count(1) \
         .block_size(Size(1, Unit.Blocks4096))\
         .oflag("direct")
+    TestRun.executor.run(f"sleep 1")
     dd.run()
     output = TestRun.executor.run(f"ls {device.system_path}* -1")
     if len(output.stdout.split('\n')) > 1:


test_tools/disk_utils.py
```diff
@@ -232,6 +232,7 @@ def remove_partitions(device):
         .count(1) \
         .block_size(Size(1, Unit.Blocks4096))\
         .oflag("direct")
+    TestRun.executor.run(f"sleep 1")
```
katlapinka, line 235: I'm just curious - why is this needed here? Did partitions unmounting process not finish before running dd?

Deixx: Short comment could be added (NOT the full explanation below ;))
We've run into a problem on one particular platform and only with HDD disks. Partitions would not disappear after dd was run and it looked like it was a race condition. `ls /dev/sdx -1` had displayed either the device and its partition (in that case partition would stay in the system), or only the device, but with error that partition cannot be accessed. I don't know how exactly ls works, but it looked like some initial scan "saw" the partition, but it was not there for the proper scan.  With trial and error we concluded that `mdadm --zero-superblock` was not releasing the device in time.
Actually maybe the sleep should be moved to the end of Mdadm.zero_superblock()... Removing partitions is quite common operation (opposed to zeroing raid metadata).

Deixx (changes requested):
Move the sleep to Mdadm.zero_superbloct()

test_tools/disk_utils.py
robertbaldyga: > With trial and error we concluded that mdadm --zero-superblock was not releasing the device in time.

Whoa, it looks like a bug in `mdadm`, doesn't it?

karolinavelkaja:
Closing this PR since better resolution had been found and placed under review in ocl: https://github.com/Open-CAS/open-cas-linux/pull/597
Problem: mdadm --zero-superblock command triggers udev and before calling next command it's needed to wait until this event ends. 
Resolution: Instead of using sleep command it's better/safer to call "udevadm settle" which would wait for event to be completed and then perform removing partitions.

PR #219, opened 2020-11-25
Update device's path handling according to incoming changes in OCL

Ostrokrzew:


storage_devices/raid.py
```diff
@@ -157,7 +157,7 @@ def create(
         Mdadm.create(raid_conf, get_devices_paths_string(volume_devices))
 
         raid_link = md_dir_path + raid_conf.name
-        raid = [r for r in Mdadm.examine_result() if r["path"] == readlink(raid_link)][0]
+        raid = [r for r in Mdadm.examine_result() if r["path"] == get_by_id_link(raid_link)][0]
```
robertbaldyga, line 160: It doesn't look correct. It seem that `Mdadm.examine_result()` returns readlinked paths, not by-id symlinks.

test_tools/fs_utils.py
```diff
@@ -287,6 +289,36 @@ def readlink(link: str, options="--canonicalize-existing"):
...
+        return path
+
+    for entry in file_paths:
+        if any([entry.full_path.startswith(x) for x in blocklist]):
+            continue
+        full_path = os.path.join(dev_by_id_dir, entry.full_path)
+        if readlink(full_path) == readlink(path):
+            return full_path
+
+    raise ValueError(f'By-id device link not found for {path}')
```
robertbaldyga, line 319: It doesn't seem to be necessary. We can just require disk paths in config to be by-id links.

test_tools/mdadm.py
```diff
@@ -53,7 +53,7 @@ def detail_result(cls, raid_device_paths: str):
             lines = device_details.splitlines()
             key = "/dev/" + lines[0].rstrip(':')
             details[key] = {}
-            details[key]["path"] = readlink(key)
+            details[key]["path"] = get_by_id_link(key)
```
robertbaldyga, line 56: It doesn't seem right. Will it work with symlink? I guess this readlink was there for a reason.

robertbaldyga (changes requested):

storage_devices/raid.py
Ostrokrzew: right

test_tools/fs_utils.py
Ostrokrzew: OK, I renamed this function and modify it to only validate input (from config or disk finder).

test_tools/mdadm.py
Ostrokrzew: You're right.

test_tools/fs_utils.py
Ostrokrzew: After some investigation I realized that we need function like this for internal purposes, so I created another function. User's input is now validated, so it must be proper. Only automated creation of objects of type `Device` (and derived) uses new function named `resolve_to_by_id_link`. If I'm wrong, please tell me.

katlapinka (approved):

storage_devices/disk.py
```diff
@@ -98,7 +98,7 @@ def __init__(
         block_size,
     ):
         Device.__init__(self, path)
-        path = fs_utils.readlink(path)
+        path = fs_utils.validate_dev_path(path)
         self.device_name = path.split('/')[-1]
```
robertbaldyga, line 102: Assuming path is always `/dev/by-id` this will not work. Please use `get_device_id()` wherever `device_name` is needed.

test_tools/fs_utils.py
```diff
@@ -287,6 +289,48 @@ def readlink(link: str, options="--canonicalize-existing"):
...
+        for entry in by_id_paths:
+            if not entry.full_path.startswith(x):
+                continue
+            if path == entry:
+                return True
+    return False
+
+
+def validate_dev_path(path: str):
+    path = os.path.abspath(path)
```
robertbaldyga, line 319: Why don't we just expect path to be absolute and raise an exception if `os.path.isabs()` returns `false`? That way we could drop return value from this function, which anyway seems unnatural for a function that is meant to just check path correctness.

storage_devices/disk.py
```diff
@@ -131,7 +131,7 @@ def is_detected(self):
                 return False
             else:
                 self.device_name = serial_numbers[self.serial_number]
-                self.system_path = f"/dev/{self.device_name}"
+                self.system_path = fs_utils.validate_dev_path(self.device_name)
```
robertbaldyga, line 134: 1. This is vulnerable to path change during reboot.
2. It could be easily replaced with `realpath(self.path)` wherever it's needed.

robertbaldyga (changes requested):

katlapinka (approved):

mmichal10 (approved):

robertbaldyga (approved):

PR #221, opened 2020-11-26
Add increment offset to fio params

katlapinka:
Needs:
https://github.com/Open-CAS/test-framework/pull/222

### DIFF
--- a/test_tools/fio/fio_param.py
+++ b/test_tools/fio/fio_param.py
@@ -205,6 +205,9 @@ class FioParam(LinuxCommand):
     def offset(self, value: Size):
         return self.set_param('offset', int(value.get_value()))
 
+    def offset_increment(self, value: Size):
+        return self.set_param('offset_increment', f"{value.value}{value.unit.get_short_name()}")
+
     def percentage_random(self, value: int):
         if value <= 100:
             return self.set_param('percentage_random', value)


test_tools/fio/fio_param.py
```diff
@@ -205,6 +205,9 @@ def num_jobs(self, value: int):
     def offset(self, value: Size):
         return self.set_param('offset', int(value.get_value()))
 
+    def offset_increment(self, value: str):
```
Ostrokrzew, line 208: I know the type is only a suggestion in Python, but I think You could put here 'int' instead of 'str', like fio's manual shows.

Ostrokrzew (dismissed):

test_tools/fio/fio_param.py
katlapinka: this parameter can be used also with string value, e.g. "50M"

Ostrokrzew: You're right, my fault, didn't check manual well.

robertbaldyga: Binary prefix units (e.g "50MiB") are also supported, so maybe we want to use `Size` instead?

robertbaldyga (changes requested):

Ostrokrzew (approved):

Ostrokrzew (approved):

robertbaldyga (approved):

PR #229, opened 2021-01-12
Add functionality to create symbolic links

Ostrokrzew:
Add functionality in `fs_utils` to create symbolic links as suggested.

### DIFF
--- a/test_tools/fs_utils.py
+++ b/test_tools/fs_utils.py
@@ -62,6 +62,10 @@ def check_if_file_exists(path):
     return TestRun.executor.run(f"test -e {path}").exit_code == 0
 
 
+def check_if_symlink_exists(path):
+    return TestRun.executor.run(f"test -L {path}").exit_code == 0
+
+
 def copy(source: str,
          destination: str,
          force: bool = False,
--- a/test_utils/filesystem/symlink.py
+++ b/test_utils/filesystem/symlink.py
@@ -2,8 +2,13 @@
 # Copyright(c) 2019-2021 Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 #
+import os
+
 from core.test_run import TestRun
-from test_tools.fs_utils import readlink
+from test_tools.fs_utils import (
+    readlink, create_directory, check_if_file_exists,
+    check_if_symlink_exists, check_if_directory_exists
+)
 from test_utils.filesystem.file import File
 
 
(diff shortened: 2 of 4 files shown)


test_tools/fs_utils.py
```diff
@@ -288,6 +288,12 @@ def readlink(link: str, options="--canonicalize-existing"):
     ).stdout
 
 
+def create_symlink(link_name, path, target):
+    if not check_if_directory_exists(path):
+        create_directory(path, True)
+    TestRun.executor.run(f"ln -s {target} {os.path.join(path, link_name)}")
+
+
```
katlapinka, line 296: I wonder if it wouldn't be better to place this method as a static method in Symlink class

Ostrokrzew: I thought about it before and I wasn't entirely convinced, but now I am.

test_utils/filesystem/symlink.py
```diff
@@ -21,3 +23,11 @@ def md5sum(self, binary=True):
 
     def get_target(self):
         return readlink(self.full_path)
+
+    @staticmethod
+    def new_symlink(link_name, path, target, force: bool = False):
+        if not check_if_directory_exists(path):
+            create_directory(path, True)
+        full_path = os.path.join(path, link_name)
+        TestRun.executor.run(f"ln -s {target} {full_path} {'-f' if force else ''}")
```
Deixx, line 32: We need a successful execution to return Symlink, so:
```suggestion
        TestRun.executor.run_expect_success(f"ln -s {target} {full_path} {'-f' if force else ''}")
```

Deixx (changes requested):

Deixx (approved):

karolinavelkaja (approved):

rafalste (approved):

test_utils/filesystem/symlink.py
```diff
@@ -21,3 +26,31 @@ def md5sum(self, binary=True):
 
     def get_target(self):
         return readlink(self.full_path)
+
+    @staticmethod
+    def get_symlink(link_path, target, **params):
```
robertbaldyga, line 31: This function essentially does not just "get" symlink. It creates it or returns existing one if present - behavior very similar to O_CREAT flag. Please rename it to `create_symlink()`.

test_utils/filesystem/symlink.py
```diff
@@ -21,3 +26,31 @@ def md5sum(self, binary=True):
...
+        if check_if_directory_exists(link_path):
+            link_path = os.path.join(link_path, Symlink.get_name(target))
+        else:
+            parent_dir = Symlink.get_parent_dir(link_path)
+            if not check_if_directory_exists(parent_dir):
+                create_directory(parent_dir, True)
```
robertbaldyga, line 37: Why does this function create a directory? It doesn't seem to be what user would expect from a function operating on symlinks.

robertbaldyga (changes requested):

test_utils/filesystem/symlink.py
karolinavelkaja: To create a symlink the specific directory must exist otherwise we get failure for link creation (No such file or directory). I think it's worth to take care of this here and I would stay with creating chosen directory.
Other solution would be to have 2 functions - one which can return failure in case of directory absence and second "don't worry, if you want this symlink I make sure you get it" like here.

karolinavelkaja:
Closing since Sławek is no longer in our team.
New review: https://github.com/Open-CAS/test-framework/pull/259

test_utils/filesystem/symlink.py
karolinavelkaja: @robertbaldyga please respond/review in new PR: https://github.com/Open-CAS/test-framework/pull/259

PR #235, opened 2021-01-25
Utility for asynchronous execution of non-async functions

Deixx:
Tests will require package added in https://github.com/Open-CAS/open-cas-linux/pull/651


Deixx:
Example:
```py
import time
from test_utils.asynchronous import start_async_func

def return_array(content, count):
    time.sleep(5)
    return [content] * count

@pytest.mark.asyncio
async def test_example():
    future_object = start_async_func(return_array, "foo", 5)
    print("process started")
    second_element = (await future_object)[1]
    print("process finished: " + str(second_element))

katlapinka (approved):

Ostrokrzew (approved):

PR #241, opened 2021-03-03
Adding functions for getting device io max size and hw io size.

karolinavelkaja:
adding functions for getting device io max size and hw io size. Also returning Size() from property.

### DIFF
--- a/storage_devices/device.py
+++ b/storage_devices/device.py
@@ -80,6 +80,12 @@ class Device:
         self.set_sysfs_property("max_sectors_kb",
                                 int(new_max_io_size.get_value(Unit.KibiByte)))
 
+    def get_max_io_size(self):
+        return Size(int(self.get_sysfs_property("max_sectors_kb")), Unit.KibiByte)
+
+    def get_max_hw_io_size(self):
+        return Size(int(self.get_sysfs_property("max_hw_sectors_kb")), Unit.KibiByte)
+
     def get_discard_granularity(self):
         return self.get_sysfs_property("discard_granularity")
 


Ostrokrzew (dismissed):

storage_devices/device.py
```diff
@@ -67,7 +67,7 @@ def get_io_stats(self):
 
     def get_sysfs_property(self, property_name):
         path = os.path.join(disk_utils.get_sysfs_path(self.get_device_id()), "queue", property_name)
-        return TestRun.executor.run_expect_success(f"cat {path}").stdout
+        return Size(int(TestRun.executor.run_expect_success(f"cat {path}").stdout), Unit.KiloByte)
```
Deixx, line 70: Sysfs properties are not only sizes, you cannot do this for all of them

Deixx (changes requested):

storage_devices/device.py
Deixx: Additionally those are more likely kibibytes not kilobytes

karolinavelkaja: Correct, will fix this! Thanks!

storage_devices/device.py
```diff
@@ -67,7 +67,7 @@ def get_io_stats(self):
 
     def get_sysfs_property(self, property_name):
         path = os.path.join(disk_utils.get_sysfs_path(self.get_device_id()), "queue", property_name)
-        return TestRun.executor.run_expect_success(f"cat {path}").stdout
+        return int(TestRun.executor.run_expect_success(f"cat {path}").stdout)
```
Ostrokrzew, line 70: You cannot cast output to `int` here - not all sysfs properties are numbers.

Ostrokrzew (changes requested):

storage_devices/device.py
karolinavelkaja: Oh ok, I've checked the usages of all properties here and I thought that all were numbers but maybe some of them which I've got zeroed may be set to other values. Thank you, I will fix this! :)

Ostrokrzew (approved):

Deixx (approved):

katlapinka (approved):

PR #243, opened 2021-03-09
add supporting methods for LVM tests

karolinavelkaja:
add supporting methods for lvm tests in new file 'lvm_common_utils.py'


Ostrokrzew (commented):
These functions are rather useful for lvm tests only, so I suggest to move them to directory with this type of tests. You can group lvm tests in fault injections sub-directory. You should consult it with one of our validation experts or maintainer.

PR #245, opened 2021-03-24
Fix for getting serial number of several disks

katlapinka:

### DIFF
--- a/test_utils/disk_finder.py
+++ b/test_utils/disk_finder.py
@@ -110,7 +110,7 @@ def get_disk_serial_number(dev_path):
     for command in commands:
         serial = TestRun.executor.run(command).stdout
         if serial:
-            return serial
+            return serial.split('\n')[0]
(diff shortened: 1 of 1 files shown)


Ostrokrzew (dismissed):

test_utils/disk_finder.py
```diff
@@ -110,6 +110,9 @@ def get_disk_serial_number(dev_path):
     for command in commands:
         serial = TestRun.executor.run(command).stdout
         if serial:
+            split_serial = serial.split('\n')
+            if len(split_serial) > 1:
+                serial = split_serial[0]
```
robertbaldyga, line 115: `split()` always returns a list (even if it has only one element) so you can do just
```python
return serial.split('\n')[0]
```

robertbaldyga (approved):

Ostrokrzew (approved):

PR #249, opened 2021-06-23
Added RamDisk API

Deixx:


Ostrokrzew (approved):
Don't we need a method to remove Ramdisks?

Deixx:
> Don't we need a method to remove Ramdisks?

They disappear after reboots or when brd module is unloaded. Do we need to specifically remove them? Unless you create an unreasonable amount of them, then the RAM usage is negligible. Otherwise they don't interfere with OS operations.

karolinavelkaja (approved):

test_utils/filesystem/symlink.py
```diff
@@ -5,18 +5,26 @@
...
 
 
 class Symlink(File):
     def __init__(self, full_path):
         File.__init__(self, full_path)
 
+    @classmethod
+    def create(cls, target_path, link_path):
+        try:
+            TestRun.executor.run_expect_success(f"ln -s {target_path} {link_path}")
```
robertbaldyga, line 18: Don't we have symlink API for that?

Btw doesn't it stem from OCL requirements? Maybe it should be part of OCL repo then? We have there bunch of OCL-oriented API anyway.

robertbaldyga (changes requested):

test_utils/filesystem/symlink.py
Deixx: This is a simple solution in the symlink API, added before realizing that there is another patch under review (https://github.com/Open-CAS/test-framework/pull/229) that does a similar thing. I've added the RFC tag to mark that the patch was not ready until we further discuss the repo scope.

PR #250, opened 2021-07-28
get_parameter_value - return the value instead of list

karolinavelkaja:


Deixx (changes requested):
Parameter values come in different shapes and sizes - for some parameters there are multiple values and thy are not necessarily numbers.

Deixx (approved):

jfckm (approved):

PR #256, opened 2021-08-16
Metods to format disks

anikielx:
Signed-off-by: Adriana Nikelska <adrianax.nikielska@intel.com>


oshchirs (changes requested):
those two functions are only nvme related, it would be better to call them from NvmeDisk class then.
please create a separate module nvme_cli.py in the test_tools folder and move those two functions there. Then add two appropriate methods to NvmeDisk which will call them.

also, the implementation of those two methods is not fully correct.
get_format_disk shall return lba format number, metadata size, and lbads (block size). Relative Performance (RP) is out of our interest. Also, it would be better to return block size in bytes.
To format_disk, in its turn, we may pass only one of two parameters - lbaf number or block size.
-f shall be turned on by default
get_format_disk I would also rename to something like "get_lba_format"
To get more info see https://nvmexpress.org/wp-content/uploads/NVM_Express_Revision_1.3.pdf Figure 115

test_tools/nvme_cli.py
```diff
@@ -0,0 +1,48 @@
+from core.test_run import TestRun
```
oshchirs, line 1: please add license

test_tools/nvme_cli.py
```diff
@@ -0,0 +1,48 @@
...
+            TestRun.LOGGER.info(
+                f"Formatting device: "
+                f"nvme format {device.path} -b {block_size} {force_param} {format_params}")
+            TestRun.executor.run_expect_success(
+                f"nvme format {device.path} -b {block_size} {force_param} {format_params}")
+            TestRun.LOGGER.info(f"Successfully format device: {device.path}")
+        else:
+            raise Exception(f"Wrong parameters to format device: {device.path}")
+    else:
+        raise Exception(f"Cannot format device without specified parameters")
```
oshchirs, line 35: no need in "f" modifier

test_tools/nvme_cli.py
```diff
@@ -0,0 +1,48 @@
...
+        lbafs = [lbaf for lbaf in lbafs if lbaf['metadata_size'] == metadata_size]
+        if block_size:
+            lbafs = [lbaf for lbaf in lbafs if lbaf['block_size'] == block_size]
+        if len(lbafs) == 1:
+            TestRun.LOGGER.info(
+                f"Formatting device: "
+                f"nvme format {device.path} -l {lbafs[0]['lba_format']} "
+                f"{force_param} {format_params}")
+            TestRun.executor.run_expect_success(
+                f"nvme format {device.path} -l {lbafs[0]['lba_format']} "
```
oshchirs, line 18: please add reset controller option

test_tools/nvme_cli.py
```diff
@@ -0,0 +1,48 @@
...
+        else:
+            raise Exception(f"Wrong parameters to format device: {device.path}")
+    elif block_size:
+        lbafs = [lbaf for lbaf in lbafs if lbaf['block_size'] == block_size]
+        if len(lbafs) > 0:
+            TestRun.LOGGER.info(
+                f"Formatting device: "
+                f"nvme format {device.path} -b {block_size} {force_param} {format_params}")
+            TestRun.executor.run_expect_success(
+                f"nvme format {device.path} -b {block_size} {force_param} {format_params}")
```
oshchirs, line 30: please add reset controller option

test_tools/nvme_cli.py
```diff
@@ -0,0 +1,48 @@
...
+        else:
+            raise Exception(f"Wrong parameters to format device: {device.path}")
+    else:
+        raise Exception(f"Cannot format device without specified parameters")
+
+
+def get_lba_format(device, in_use_param=False):
+    in_use_param = "| grep '(in use)'" if in_use_param else ''
+    lines = TestRun.executor.run_expect_success(
+        f"nvme id-ns {device.path} | grep lbads {in_use_param}").stdout.splitlines()
```
oshchirs, line 41: please use json output format, so you won't need to parse cli output manually using split and indexes
instead of "in use" you can use flbas field (use bits 3:0)

test_tools/nvme_cli.py
```diff
@@ -0,0 +1,48 @@
...
+def format_disk(device, metadata_size=None, block_size=None, force=True, format_params=None):
+    force_param = '-f' if force else ''
+    format_params = ' '.join(format_params) if format_params else ''
+    lbafs = get_lba_format(device)
+    if metadata_size:
+        lbafs = [lbaf for lbaf in lbafs if lbaf['metadata_size'] == metadata_size]
+        if block_size:
+            lbafs = [lbaf for lbaf in lbafs if lbaf['block_size'] == block_size]
+        if len(lbafs) == 1:
+            TestRun.LOGGER.info(
```
oshchirs, line 13: No need to write the whole command in the log.
Just print smth like "Formatting /dev/nvme1n1 device with 4096 block size and 64 metadata size"

test_tools/nvme_cli.py
```diff
@@ -0,0 +1,48 @@
...
+                f"nvme format {device.path} -l {lbafs[0]['lba_format']} "
+                f"{force_param} {format_params}")
+            TestRun.LOGGER.info(f"Successfully format device: {device.path}")
+        else:
+            raise Exception(f"Wrong parameters to format device: {device.path}")
+    elif block_size:
+        lbafs = [lbaf for lbaf in lbafs if lbaf['block_size'] == block_size]
+        if len(lbafs) > 0:
+            TestRun.LOGGER.info(
+                f"Formatting device: "
```
oshchirs, line 27: No need to write the whole command in the log.
Just print smth like "Formatting /dev/nvme1n1 device with 4096 block size and 64 metadata size"

oshchirs (changes requested):
overall looks good. added few comments

Deixx (approved):

katlapinka (approved):

PR #257, opened 2021-08-18
Add PCI address to NvmeDisk object

oshchirs:
Signed-off-by: Oleksandr Shchirskyi <oleksand.shchirskyi@intel.com>

### DIFF
--- a/storage_devices/device.py
+++ b/storage_devices/device.py
@@ -6,7 +6,7 @@ import os
 
 from core.test_run import TestRun
 from test_tools import disk_utils, fs_utils
-from test_tools.disk_utils import get_device_filesystem_type
+from test_tools.disk_utils import get_device_filesystem_type, get_pci_address
 from test_utils.io_stats import IoStats
 from test_utils.size import Size, Unit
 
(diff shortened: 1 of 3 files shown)


storage_devices/device.py
```diff
@@ -17,6 +17,7 @@ def __init__(self, path):
         self.path = path
         self.size = Size(disk_utils.get_size(self.get_device_id()), Unit.Byte)
         self.filesystem = get_device_filesystem_type(self.get_device_id())
+        self.pci_address = get_pci_address(self.get_device_id())
```
katlapinka, line 20: I think this should be added to NvmeDisk class, because there are other devices which inherit from Device class and they do not have such property, i.e. Raid, ErrorDevice

katlapinka (changes requested):

storage_devices/device.py
oshchirs: done

storage_devices/disk.py
```diff
@@ -186,6 +187,7 @@ class NvmeDisk(Disk):
 
     def __init__(self, path, disk_type, serial_number, block_size):
         Disk.__init__(self, path, disk_type, serial_number, block_size)
+        self.pci_address = get_pci_address(self.get_device_id())
```
katlapinka, line 190: I wonder if it wouldn't be helpful to add __str__ method to this class, so this pci_address would be printed at the begining of log, where all disks are listed. (See __str__ method in Disk class)

Deixx (dismissed):

storage_devices/disk.py
oshchirs: done

katlapinka (approved):

Deixx (approved):

PR #258, opened 2021-08-19
Add possibility to specify extra logs to dump

oshchirs:
Signed-off-by: Oleksandr Shchirskyi <oleksand.shchirskyi@intel.com>

### DIFF
--- a/log/logger.py
+++ b/log/logger.py
@@ -182,8 +182,10 @@ class Log(HtmlLogManager, metaclass=Singleton):
         if not check_if_file_exists(messages_log):
             messages_log = "/var/log/syslog"
         log_files = {"messages": messages_log,
-                     "dmesg": "/tmp/dmesg",
-                     "cas": "/var/log/opencas.log"}
+                     "dmesg": "/tmp/dmesg"}
+        if hasattr(TestRun.usr, "logs_to_dump"):
+            log_files.update(TestRun.usr.logs_to_dump)
+
         TestRun.executor.run(f"dmesg > {log_files['dmesg']}")
 
         for log_name, log_source_path in log_files.items():
@@ -196,7 +198,6 @@ class Log(HtmlLogManager, metaclass=Singleton):
 
     def generate_summary(self, item, meta):
         import json
-        import hashlib
         summary_path = os.path.join(self.base_dir, 'info.json')
         with open(summary_path, "w+") as summary:
             data = {
--- a/log/logger.py
+++ b/log/logger.py
@@ -183,8 +183,8 @@ class Log(HtmlLogManager, metaclass=Singleton):
             messages_log = "/var/log/syslog"
         log_files = {"messages": messages_log,
                      "dmesg": "/tmp/dmesg"}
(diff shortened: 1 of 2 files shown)


katlapinka (dismissed):

log/logger.py
```diff
@@ -182,8 +182,10 @@ def get_additional_logs(self):
         if not check_if_file_exists(messages_log):
             messages_log = "/var/log/syslog"
         log_files = {"messages": messages_log,
-                     "dmesg": "/tmp/dmesg",
-                     "cas": "/var/log/opencas.log"}
+                     "dmesg": "/tmp/dmesg"}
+        if hasattr(TestRun.usr, "logs_to_dump"):
+            log_files.update(TestRun.usr.logs_to_dump)
```
robertbaldyga, line 187: test-framework should never access `TestRun.usr`. This is private field reserved for TF user. Please use `dut_config` instead to pass list of paths. (You can see how plugin configuration is handled. The list of logs may be specified in `conftest.py` before `TestRun.prepare()` is called).

robertbaldyga (changes requested):

log/logger.py
oshchirs: done

Deixx (approved):
I'm wondering if we should make a distinction between logs, which can be easily copied from a platform (i.e. `/var/log/messages`) and those that require dumping a command output first (i.e. `dmesg`, `cat /proc/mdstat`).
Sth like:
```
logs = { "messages": "/var/log/messages",
         "dmesg": "dmesg"
         "mdstat": "cat /proc/mdstat"
}

for log_name, log_source_path in log_files.items():
    if not file_exists(log_source_path):    # will not exist for "command logs"
        executor.run(f"{log_source_path} > /tmp/{log_name})
        log_source_path = "/tmp/" + log_name
    ...
    executor.rsync_from(log_source_path, log_destination_path)
```

katlapinka (approved):

robertbaldyga (approved):

robertbaldyga:
> I'm wondering if we should make a distinction between logs, which can be easily copied from a platform (i.e. `/var/log/messages`) and those that require dumping a command output first (i.e. `dmesg`, `cat /proc/mdstat`).
> Sth like:
> 
> ```
> logs = { "messages": "/var/log/messages",
>          "dmesg": "dmesg"
>          "mdstat": "cat /proc/mdstat"
> }
> 
> for log_name, log_source_path in log_files.items():
>     if not file_exists(log_source_path):    # will not exist for "command logs"
>         executor.run(f"{log_source_path} > /tmp/{log_name})
>         log_source_path = "/tmp/" + log_name
>     ...
>     executor.rsync_from(log_source_path, log_destination_path)
> ```

This looks a bit fragile. We could make full-fledged API for that instead. Something like
```python
TestRun.attach_log(log_path, target_name=None, dump_fn=None)
```
where `dump_fn` would be a function that dumps the actual log to a file. That way we can even have logs that require more sophisticated extraction method, or logs that are filtered or preprocessed in some other way.

PR #259, opened 2021-08-20
Add functionality to create symbolic links

karolinavelkaja:
Original review: https://github.com/Open-CAS/test-framework/pull/229


test_utils/filesystem/symlink.py
```diff
@@ -21,3 +26,30 @@ def md5sum(self, binary=True):
...
+                        return Symlink(link_path)
+                    else:
+                        raise FileExistsError("Symbolic link already exists and leads to the other "
+                                              "target.")
+                else:
+                    cmd = f"ln -s {target} {link_path}"
+                    for param, value in params.items():
+                        cmd += f' --{str(param)} {str(value)}'
+                    TestRun.executor.run_expect_success(cmd)
+                    return Symlink(link_path)
```
rafalste, line 55: This part of code is indented too much, resulting with no action if _parent_dir_ exists.

test_utils/filesystem/symlink.py
```diff
@@ -21,3 +26,30 @@ def md5sum(self, binary=True):
 
     def get_target(self):
         return readlink(self.full_path)
+
+    @staticmethod
+    def create_symlink(link_path, target, **params):
```
rafalste, line 31: Maybe switch places of _link_path_ and _target_ to be more inline with `ln` argument order?

test_utils/filesystem/symlink.py
```diff
@@ -21,3 +26,30 @@ def md5sum(self, binary=True):
...
+        else:
+            parent_dir = Symlink.get_parent_dir(link_path)
+            if not check_if_directory_exists(parent_dir):
+                create_directory(parent_dir, True)
+                file_exists = check_if_file_exists(link_path)
+                is_symlink = check_if_symlink_exists(link_path)
+                if file_exists and not is_symlink:
+                    raise FileExistsError("Overwriting item that is not symbolic link could cause "
+                                          "problems and will not be executed.")
+                elif file_exists and is_symlink and not bool(params.get('force')):
```
rafalste, line 44: `not bool(params.get('force'))` will return True even if there is a _force_ parameter but without value (i.e. value set to None, 0 or empty string) and _force_ doesn't need any value, so I assume there will be none. That is because `dict.get(name)` returns value of given name.

Better way of checking the absence of _force_ parameter here, would be to put it simply like this:
`elif file_exists and is_symlink and 'force' not in params:`
as this checks names rather then values.

rafalste: Ok, now I can see that each parameter probably needs to be assigned with at least _True_ value. If I'm on the right track now and this is the case, then the other part of code in line 53 needs to be modified to not pass a parameter with a value in form of `--force True` as this will not work.

robertbaldyga (changes requested):

PR #266, opened 2021-09-09
Update fio version

katlapinka:


karolinavelkaja (approved):

Deixx:
Is there a motivation for this update?

katlapinka (commented):
Older fio versions have some bug because of which they consume too large amount of RAM in some scenarios. The problem is not observed in this version

Deixx (approved):

PR #271, opened 2021-10-07
Add random_generator param to fio

katlapinka:


karolinavelkaja (approved):

Deixx (changes requested):
I prefer the version with the enum

